PR #2, opened 2023-01-30
Update HttpChangeConsumer.java - Adding ability to attach autorization header

yanivbh1:
Adding ability to add authorization header
to support JWT-type authentication which is highly common pattern and will enable variety of use cases with debezium and HTTP sink


jpechane:
@yanivbh1 Could you please create a Jira issue for this?
Also I beleive the problem needs to be handled in more general way

* Debezium Server now supports sinks the ability to use headers passed - see Kafka sink for implementation. The code sould support it
* Instead ofconfiguring http headers one by one there should be a set `header.*` config options that will take the last part and addsuch header to request. This will enable user configure any header
* In case of coflict of two items above the incoming header takes precedence and overwirtes the one statically configured

jpechane:
@gunnarmorling I think it still is. So far you can either use jwt authentication or pass Kafka Connect headers. But there is no way how to configure an HTTP header statically in the Debezium Server config.

Naros:
:x: Developer Certificate of Origin (DCO) check failed.

Hi @yanivbh1, please sign off all commits with:

```
git commit -s
```

If pull request commits are not signed off, the pull request cannot be merged. For more information about why this is required, please see [our blog](https://debezium.io/blog/2025/12/12/contribution-requirements-changing/) about contribution requirement changes.

PR #6, opened 2023-02-05
DBZ-6017 Add support for header to all Debezium Server sinks

jeremy-l-ford:
Add header support to http sink


jpechane:
@jeremy-l-ford Nice work, applied, thanks!

PR #8, opened 2023-02-20
DBZ-6129 Upgrade to Quarkus 3.0.0.Final

Naros:
https://issues.redhat.com/browse/DBZ-6129

A sibling PR for https://github.com/debezium/debezium/pull/4299, which upgrades this repository to Quarkus 3.  This PR should not be merged until after the parent repository's PR has been.

### DIFF
--- a/debezium-server-core/pom.xml
+++ b/debezium-server-core/pom.xml
@@ -81,6 +81,12 @@
             <groupId>io.confluent</groupId>
             <artifactId>kafka-connect-protobuf-converter</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.protobuf</groupId>
+                    <artifactId>protobuf-java</artifactId>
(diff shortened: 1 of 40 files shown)


Naros:
@ani-sha should this PR not have checked out the PR in the main repository associated with DBZ-6129 when doing the cross maven build?  It seems it used the default branch.

Naros:
So semi-related to this PR, while I was testing these changes I noticed that the Redis tests fail regularly locally.  While looking into why, I found that if I instead use a `FixedHostPortGenericContainer` and export Redis on either port `6739` or `16739`, the tests successfully run every time.  Normally Testcontainers was assigning ports in the range of 30000+ and this was always failing, so I need to dig deeper as I don't think we can afford to explicitly pick a port and run the tests on that port, can we @jpechane ?

Naros:
The test failures with Redis is unrelated.

PR #9, opened 2023-02-20
DBZ-6133 Attempt at fixing Redis test failures

Naros:


Naros:
Hi @vjuranek and @jpechane, I am curious whether this is pure luck or if there is some technical difference here using this method to obtain the port number, but the tests all passed.  Do we want to push this to the main branch and see if it keeps Redis stable?

vjuranek (approved):
The only exposed port (also in [Dockerfile](https://github.com/docker-library/redis/blob/master/7.0/Dockerfile#L119)) is `6379`, so IMHO both should return the same port. If it fails randomly (e.g. for me the tests are locally passing) with `getFirstMappedPort()`, IMHO it's probably some bug in Testcontainers (maybe caused by different Docker version?). Anyway, `getMappedPort()` seems to be more correct as we actually don't want first mapped port, but mapping of port `6379`, so +1 from me for this change.

jpechane:
@Naros Applied, thanks. Let's give it a try. As menitoned this might be bug or something so unless we start using it we will not be sure.

PR #15, opened 2023-04-07
DBZ-6319 Async producer for Pulsar sink

henkosch:
We have experienced very low performance using the current pulsar sink implementation. The root cause seems to be that it uses the synchronous `send` method and it waits for every single message to be delivered and written to pulsar before trying to send the next one. 

For much better message rate it is recommended to use `sendAsync` for the items in the batch and wait all of them to complete before marking the batch finished.

Real debezium performance measurements:
* sending messages one-by-one using `send` on my m1 macbook pro: `~600 msg/sec`
* sending messages in batches using `sendAsync` on my m1 macbook pro: `~70000 msg/sec`

Measured reate using pulsar-perf cli tool.

### DIFF
--- a/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
+++ b/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
@@ -148,21 +148,22 @@ public class PulsarChangeConsumer extends BaseChangeConsumer implements Debezium
 
             final ChangeEvent<Object, Object> currentRecord = record;
 
-            CompletableFuture<MessageId> future = message.sendAsync()
-                    .thenAccept(messageId -> {
-                        LOGGER.trace("Sent message with id: {}", messageId);
-                        try {
-                            committer.markProcessed(currentRecord);
-                        }
-                        catch (InterruptedException e) {
-                            throw new DebeziumException(e);
-                        }
-                    })
-                    .exceptionally(ex -> {
-                        Throwable exception = (Throwable) ex;
-                        LOGGER.error("Failed to send record to {}:", record.destination(), exception);
-                        throw new DebeziumException(exception);
-                    });
+            CompletableFuture<MessageId> future = message
+                .sendAsync()
+                .thenAccept(messageId -> {
+                    LOGGER.trace("Sent message with id: {}", messageId);
+                    try {
+                        committer.markProcessed(currentRecord);
+                    }
+                    catch (InterruptedException e) {
+                        throw new DebeziumException(e);
+                    }
+                })
+                .exceptionally(ex -> {
+                    Throwable exception = (Throwable) ex;
+                    LOGGER.error("Failed to send record to {}:", record.destination(), exception);
+                    throw new DebeziumException(exception);
+                });
 
             futures.add(future);
         }
@@ -170,8 +171,8 @@ public class PulsarChangeConsumer extends BaseChangeConsumer implements Debezium
         try {
             // Wait for all CompletableFutures to complete
             CompletableFuture
-                    .allOf(futures.toArray(new CompletableFuture[0]))
-                    .join();
+                .allOf(futures.toArray(new CompletableFuture[0]))
+                .join();
         }
         catch (CompletionException e) {
             if (e.getCause() instanceof DebeziumException) {
--- a/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
+++ b/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
@@ -175,17 +175,21 @@ public class PulsarChangeConsumer extends BaseChangeConsumer implements Debezium
 
         // Flush all producers asynchronously
         // Waiting for the returned futures will wait until all messages have been successfully persisted.
-        CompletableFuture<Void>[] futures = batchProducers
-                .values()
-                .stream()
-                .map(Producer::flushAsync)
-                .toArray(CompletableFuture[]::new);
+        CompletableFuture<Void> allProducersCompleted = CompletableFuture
+                .allOf(batchProducers
+                        .values()
+                        .stream()
+                        .map(Producer::flushAsync)
+                        .toArray(CompletableFuture[]::new));
 
         try {
-            // Wait for all producers to complete the flush
-            CompletableFuture
-                    .allOf(futures)
-                    .get(timeout, TimeUnit.MILLISECONDS);
+            // Wait for all producers to complete the flush with timeout
+            if (timeout > 0) {
+                allProducersCompleted.get(timeout, TimeUnit.MILLISECONDS);
+            }
+            else {
+                allProducersCompleted.join();
+            }
         }
         catch (CompletionException | ExecutionException | TimeoutException exception) {
             LOGGER.error("Failed to send batch:", exception);
--- a/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
+++ b/debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
@@ -11,14 +11,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
-import java.util.concurrent.CountDownLatch;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.enterprise.context.Dependent;
 import javax.inject.Named;
 
-import io.netty.util.concurrent.CompleteFuture;
 import org.apache.pulsar.client.api.MessageId;
 import org.apache.pulsar.client.api.Producer;
 import org.apache.pulsar.client.api.PulsarClient;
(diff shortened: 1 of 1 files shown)


Naros:
Hi @henkosch can you update the debezium server docs to include the new configuration option?

henkosch:
@Naros Sure, here are the docs for the new `async` option: https://github.com/debezium/debezium/pull/4442

Also it seems that the integration tests are failing for the Redis sync adapter for some reason, but I doubt it would have anything to do with my changes in pulsar.

```
[INFO] Reactor Summary for Debezium Server Parent 2.2.0-SNAPSHOT:
[INFO] 
[INFO] Debezium Server BOM ................................ SUCCESS [  1.185 s]
[INFO] Debezium Server Parent ............................. SUCCESS [ 19.206 s]
[INFO] Debezium Server Core ............................... SUCCESS [02:55 min]
[INFO] Debezium Server Amazon Kinesis Sink Adapter ........ SUCCESS [ 19.864 s]
[INFO] Debezium Server Google Cloud Pub/Sub Sink Adapter .. SUCCESS [ 42.501 s]
[INFO] Debezium Server Pulsar Sink Adapter ................ SUCCESS [01:12 min]
[INFO] Debezium Server Azure Event Hubs Adapter ........... SUCCESS [ 10.489 s]
[INFO] Debezium Server HTTP Webhook Sink Adapter .......... SUCCESS [ 53.130 s]
[INFO] Debezium Server Redis Sink Adapter ................. FAILURE [03:43 min]
[INFO] Debezium Server Kafka Sink Adapter ................. SUCCESS [ 39.099 s]
[INFO] Debezium Server Pravega Sink Adapter ............... SUCCESS [01:30 min]
[INFO] Debezium Server NATS Streaming Sink Adapter ........ SUCCESS [ 20.434 s]
[INFO] Debezium Server NATS JetStream Sink Adapter ........ SUCCESS [ 18.707 s]
[INFO] Debezium Server Infinispan Sink Adapter ............ SUCCESS [ 42.326 s]
[INFO] Debezium Server RabbitMQ Stream Sink Adapter ....... SUCCESS [ 28.166 s]
[INFO] Debezium Server RocketMQ Sink Adapter .............. SUCCESS [02:02 min]
[INFO] Debezium Server Distribution ....................... SKIPPED
```

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                        try {
+                            committer.markProcessed(currentRecord);
+                        }
+                        catch (InterruptedException e) {
+                            throw new DebeziumException(e);
+                        }
+                    })
+                    .exceptionally(ex -> {
+                        Throwable exception = (Throwable) ex;
+                        LOGGER.error("Failed to send record to {}:", record.destination(), exception);
```
codelipenghui, line 201: It will be a message-based error log.
For Pulsar, if the first message is failed to publish, all the subsequent messages will also be failed.
I think we can change the log level to trace here and add an error log between line:214 and line:215.
To make sure it will not be noisy.

codelipenghui (commented):
Thanks, @henkosch
You found the major performance bottleneck of the connector.
And I think your fix will definitely fix the issue.

I would like to share my thoughts about the new configuration that will be introduced.

IMO, the new configuration is not necessary. We should always use the async API in connectors. I can't imagine what kind of scene should use the sync API in the connector. The connector always processes a set of records and then pushes them to a topic. The async API is the more natural way to work with this case.

1. Write all the records to the pending buffer of the producer
2. Flush the buffer
3. Wait for the publish receipts from the broker of all the flushed records

The sync API doesn't looks reasonable here.

1. Get a set of records that need to send to a topic
2. Peek the first record and send it to the topic
3. Wait for the publish receipt from the broker
4. Peek the second record and send it to the topic
5. Wait for the publish receipt
6. Keep going the next record

And users will not see any different behavior with different `asyncEnabled` settings except for the difference in performance. So this is why I think we don't need a configuration to control `bad performance` and `good performance`. Everyone wants good performance.

And if we switch to async API, we are safe by using the default configuration of the batch policy (1000 messages and 128KB). What we should expose to users is the batch policy. Users can tune the configurations to get higher throughput or lower latency. We can set the following batch options in Pulsar's producer

- batchingMaxMessages (default is 1000)
- batchingMaxBytes (default is 128KB)
- batchingMaxPublishDelay (default is 1ms)

But it is not required for this PR. The default configuration can work well for most cases.
So my suggestion here to remove the way that using sync method, use the async method directly. And no breaking changes will be introduced. So we are safe here.

jpechane:
I agrre with the proposal to go async only, it will simplify the code a bit. I don't see any advantage in keeping the sync imeplementation.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+            final TypedMessageBuilder message = buildMessage(record);
+
+            final ChangeEvent<Object, Object> currentRecord = record;
+
+            CompletableFuture<MessageId> future = message
+                    .sendAsync()
+                    .thenAccept(messageId -> {
+                        LOGGER.trace("Sent message with id: {}", messageId);
+                        try {
+                            committer.markProcessed(currentRecord);
```
jpechane, line 193: I am not fully convinced on this one. We cannot expect that the recordds will be delivered in the order, can we? If not then the last committed offset might not be the latest one form the batch,

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                        throw new DebeziumException(exception);
+                    });
+
+            futures.add(future);
+        }
+
+        try {
+            // Wait for all CompletableFutures to complete
+            CompletableFuture
+                    .allOf(futures.toArray(new CompletableFuture[0]))
```
jpechane, line 211: Could we use the right size of array as we know how many futures we created?

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                    .join();
+        }
+        catch (CompletionException e) {
+            if (e.getCause() instanceof DebeziumException) {
```
jpechane, line 215: I don't think it is necessary to unwind the exception. Let's jesyt either let `CompletionException` get propagated or use `throw new DebeziumException(e)`

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                    });
+
+            futures.add(future);
+        }
+
+        try {
+            // Wait for all CompletableFutures to complete
+            CompletableFuture
+                    .allOf(futures.toArray(new CompletableFuture[0]))
+                    .join();
```
jpechane, line 212: I'd recommend to provide a time-bound solution too - see https://issues.redhat.com/browse/DBZ-5175?focusedId=20451392&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20451392 for an issue we have in Google PubSub sink so better to be ready for that too. The timeout would be configurable with default 0 meaning unlimited.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                    })
+                    .exceptionally(ex -> {
+                        Throwable exception = (Throwable) ex;
+                        LOGGER.error("Failed to send record to {}:", record.destination(), exception);
+                        throw new DebeziumException(exception);
+                    });
+
+            futures.add(future);
+        }
+
```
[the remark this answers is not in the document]
henkosch: @codelipenghui I have refactored the code to use flush instead, as you suggested. It turned out that [flushAsync](https://pulsar.apache.org/api/client/2.4.0/org/apache/pulsar/client/api/Producer.html#flushAsync--) on a producer waits for all the messages to be successfully delivered, so calling flush should be enough and we don't have to wait for individual message send futures anymore. Please take a look.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                            committer.markProcessed(currentRecord);
+                        }
+                        catch (InterruptedException e) {
+                            throw new DebeziumException(e);
+                        }
+                    })
+                    .exceptionally(ex -> {
+                        Throwable exception = (Throwable) ex;
+                        LOGGER.error("Failed to send record to {}:", record.destination(), exception);
+                        throw new DebeziumException(exception);
```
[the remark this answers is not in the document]
henkosch: @codelipenghui Using exceptionally we can handle and recover from exceptions and it was not what we wanted here, so I changed it to  whenComplete instead. Take a look at my new commit.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
```diff
@@ -152,4 +175,52 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+            CompletableFuture
+                    .allOf(futures.toArray(new CompletableFuture[0]))
+                    .join();
+        }
+        catch (CompletionException e) {
+            if (e.getCause() instanceof DebeziumException) {
+                throw (DebeziumException) e.getCause();
+            }
+            else {
+                throw new DebeziumException(e);
```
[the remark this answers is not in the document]
henkosch: Done.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
henkosch: Done.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
henkosch: Done.

debezium-server-pulsar/src/main/java/io/debezium/server/pulsar/PulsarChangeConsumer.java
henkosch: Done. I have created a new configuration property for it, called `timeout`. Is it okay this way?

henkosch:
With the new `flushAsync` based implementation I was able to measure ` 72043  msg/s` which is double of my original measurements. So great! :)

jpechane:
@henkosch Merged via https://github.com/debezium/debezium-server/pull/16
Thanks a lot
The only task left is the dos PR update.

PR #18, opened 2023-04-17
DBZ-6346 Change severity of a log message for OOM

eizners:
https://issues.redhat.com/browse/DBZ-6346


jpechane:
@eizners Applied, thanks!

PR #19, opened 2023-04-20
DBZ-6372 - Kinesis Sink - AWS Credentials Provider

brenoavm:


debezium-server-kinesis/src/main/java/io/debezium/server/kinesis/KinesisChangeConsumer.java
```diff
@@ -78,8 +78,7 @@ void connect() {
         region = config.getValue(PROP_REGION_NAME, String.class);
         endpointOverride = config.getOptionalValue(PROP_ENDPOINT_NAME, String.class);
         final KinesisClientBuilder builder = KinesisClient.builder()
-                .region(Region.of(region))
-                .credentialsProvider(ProfileCredentialsProvider.create(credentialsProfile));
+                .region(Region.of(region));
```
Naros, line 81: With the removal of `ProfileCredentialsProvider` usage here, I suspect the import above is not used, could the unused imports also be removed as well?

brenoavm: Yes, we probably can remove the unused imports. I will commit it again.

jpechane:
@brenoavm Thanks for the PR
IMHO to make sure that this feature is backward compaible we should apply a differnt approach.
Credentials profile property should be checked if it is present - instead of assuming it will be `default` if not. If present, then `ProfileCredentialsProvider` will be used, otherwise it will not. In that case it is also important to update the documentation to reflet the change.

Naros (approved):
Overall this LGTM, but could you please remove the merge commit @brenoavm ?
There should also be a sibling PR opened in the debezium/debezium repository updating the documentation accordingly.

jpechane:
@brenoavm Squased and applied, thanks!

PR #20, opened 2023-04-25
DBZ-6348 Add JWT authentication to HTTP Client

rnowling-memphis:
This PR adds a JWT authentication option to the HTTP Client sink (see [DBZ-6348](https://issues.redhat.com/browse/DBZ-6348)).  If the JWT authentication is specified, the client will have the authenticator check for an active authentication token.  If there is no token or its expired, the authenticator will make a request for an initial or refresh token.  The authentication details will then be added to the HTTP request header.

The following options are added to the config file:

* `debezium.sink.http.authentication.type`: If missing, no authentication is used.  If present, its value specifies the authentication protocol.  The only currently supported value is `jwt`.
* `debezium.sink.http.authentication.jwt.username`: Specify the username
* `debezium.sink.http.authentication.jwt.password`: Specify the password
* `debezium.sink.http.authentication.jwt.url`: Specify the base URL.  E.g., `http://myserver:8000/`.  The paths `auth/authenticate` and `auth/refreshToken` are appended for the JWT REST calls.
* `debezium.sink.http.authentication.jwt.token_expiration`: Requested duration before token expires (in minutes)
* `debezium.sink.http.authentication.jwt.refresh_token_expiration`: Requested duration before refresh token expires (in minutes)

A separate PR to the Debezium repo will be needed to add documentation.

This PR also adds instructions to the README.md for building Debezium Server or local development.

I tested this end-to-end against a HTTP service that supports JWT authorization.  Authorization was successful.  CDC events were properly sent to the endpoint using the authorization credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The application itself contains the `core` module and a set of modules responsib
 The following software is required to work with the Debezium codebase and build it locally:
 
 * [Git](https://git-scm.com) 2.2.1 or later
-* JDK 17 or later, e.g. [OpenJDK](http://openjdk.java.net/projects/jdk/)
+* JDK 11 or later, e.g. [OpenJDK](http://openjdk.java.net/projects/jdk/)
 * [Docker Engine](https://docs.docker.com/engine/install/) or [Docker Desktop](https://docs.docker.com/desktop/) 1.9 or later
 * [Apache Maven](https://maven.apache.org/index.html) 3.8.4 or later  
   (or invoke the wrapper with `./mvnw` for Maven commands)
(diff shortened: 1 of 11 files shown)


jpechane (commented):
@rnowling-memphis I pushed a commit with udated exception handling.
The code LGTM but I miss test for it. Ideally there should be an integration test but unit test will be fine as well. I noticed there are nethods maked as visible for testing but no tests are present. Did you forget to commit them?

rnowling-memphis:
I didn't add tests, but I absolutely can.  I'll pull your changes and add some unit tests on top of it.  Thanks for the review @jpechane !

jpechane:
@rnowling-memphis Applied, thanks!

PR #22, opened 2023-04-29
DBZ-6405 Set (instead of adding) Authorization Headers

rnowling-memphis:
Set (overwrite if exists) authorization headers rather than adding them (resulting in duplicates).  Fixes [DBZ-6405](https://issues.redhat.com/browse/DBZ-6405).


Naros:
Applied to main (=2.3).

PR #23, opened 2023-05-15
DBZ-6461 Debezium Server cannot recover from Google Pub/Sub errors

mfvitale:
Add properties 
`debezium.sink.pubsub.wait.message.computation.timeout.ms` and `debezium.sink.pubsublite.wait.message.computation.timeout.ms`
 to configure the wait time on message processing by Google PubSub. In this wait the connector can fail and be restarted.
 
 Doc PR https://github.com/debezium/debezium/pull/4544

### DIFF
--- a/debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
+++ b/debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
@@ -125,8 +125,8 @@ public class PubSubChangeConsumer extends BaseChangeConsumer implements Debezium
     @ConfigProperty(name = PROP_PREFIX + "retry.rpc.timeout.multiplier", defaultValue = "2.0")
     Double rpcTimeoutMultiplier;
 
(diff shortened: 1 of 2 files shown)


debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
```diff
@@ -123,6 +125,9 @@ public interface PublisherBuilder {
     @ConfigProperty(name = PROP_PREFIX + "retry.rpc.timeout.multiplier", defaultValue = "2.0")
     Double rpcTimeoutMultiplier;
 
+    @ConfigProperty(name = PROP_PREFIX + "wait.message.computation.timeout.ms", defaultValue = "5000")
```
jpechane, line 128: @mfvitale We are going from unlimited. It is really hard to know what is the usual time till the mesage batch is written.
Is not 5 secs value a bit too aggressive? I'd prefer to err on the safe side at least in the beginning and choose 30 or 60 secs.
Also why `computation`? Would not `delivery` be better to represent the value use?

jpechane:
@mfvitale Applied, thanks

PR #25, opened 2023-05-29
DBZ-6493 Adding kubernetes-config extension in order to allow configuration from k8 config maps and secrets

jcechace:
This is part of initial implementation of Debezium Operator 
https://issues.redhat.com/browse/DBZ-6493

Safely storing credentials is vital for proper k8 support.

### DIFF
--- a/debezium-server-dist/pom.xml
+++ b/debezium-server-dist/pom.xml
@@ -29,6 +29,11 @@
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
                 <version>${quarkus.version}</version>
+                <configuration>
+                    <systemProperties>
+                        <quarkus.kubernetes-config.secrets.enabled>true</quarkus.kubernetes-config.secrets.enabled>
+                    </systemProperties>
+                </configuration>
                 <executions>
                     <execution>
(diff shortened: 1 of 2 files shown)


debezium-server-dist/pom.xml
```diff
@@ -29,6 +29,11 @@
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
                 <version>${quarkus.version}</version>
+                <configuration>
+                    <systemProperties>
+                        <quarkus.kubernetes-config.secrets.enabled>true</quarkus.kubernetes-config.secrets.enabled>
```
jcechace, line 34: See: https://quarkus.io/guides/kubernetes-config#quarkus-kubernetes-config_quarkus.kubernetes-config.secrets.enabled

This needs to be here to enable secret support. However, since we are not using the `kubernetes` extension no resources should be generated and the entire support for reading config values from k8 should be controlled by runtime property `

This is based on my observation and on https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/secret.20value.20with.20kubernetes-config.20extension/near/361962181


However I do find the documentation for the build time property confusing

PR #28, opened 2023-06-19
DBZ-6574 Disabling k8 dev service by default to not interfere with SSL tests

jcechace:
https://issues.redhat.com/browse/DBZ-6574


jpechane:
@jcechace LGTM, thanks!

PR #31, opened 2023-07-06
Dbz 6638 multiple cassandra versions

eizners:


jpechane:
@eizners Cassandra 3 contains 250 MB of deps, v4 and DSE looks correct

jpechane:
@eizners This should be done, please review the change and test it.
The usage is like
`env EXTRA_CONNECTOR=cassandra-3 ./run.sh` or `env EXTRA_CONNECTOR=dse ./run.sh`
The package file is now clean with only few duplicate files present

eizners:
Made fixes and tested DSE and Cassandra4.

jpechane:
@eizners Removed merge commit and applied, thanks for the good job!

jpechane:
We just need to update Debezium Server docs.

PR #32, opened 2023-07-24
DBZ-6701 enabling JNDI in order to support SRV protocol in MongoDB connector

jcechace:
https://issues.redhat.com/browse/DBZ-6701


jcechace:
Should be also backported to 2.3

jpechane:
@jcechace Applied, thanks

PR #33, opened 2023-08-01
DBZ-6723 Add partitioning support for Azure EventHubs sink

slknijnenburg:
## Background

With the existing Azure EventHubs sink implementation it is not possible to leverage multiple partitions on the EventHubs side. One can either hardcode the partitionID where _all_ messages will be sent to, or a fixed partitionKey, which will make EventHubs pick a single partition derived from that key. Being only able to leverage a single partition makes it hard to parallelise any downstream processing. This is something we ran into when processing ~900GB of messages from a single large database server.

The EventHubs SDK only allows specifying the partition (either via an ID or a key) on a batch level, as the entire batch of messages is being emitted as a single message under the hood. This means that in order to send messages to N partitions, we need to manage N potential batches and we need to route  events to the correct batch to ultimately make the message arrive in the correct EventHubs partition.

## Changes

In this PR I've implemented the following changes:
- Automatically detect the number of partitions in the connected EventHub
- Route events to a given partitionID.
- Manages N batches for N partitions when processing events
- Only creates batches when events are actually being routed to it, preventing the overhead of opening up N connections for N partitions in parallel.

The changeset is intended to be fully backwards compatible. Two new configuration properties were added: 
- `partitioning.selector`: Chooses which part of the `record` is used to derive the partition from. Must be `destination`, `key`, or `value`. 
- `partitioning.field`: Allows to configure a dot-notation field that is queried on the record's key or value. This field is read, it's hashcode is modulo'd against the number of partitions to find the actual partitionID where the event should be routed to.

The new properties are ignored if previous configuration for `partitionid` or `partitionkey` are present.

## Other notes

- The partitionID assignment is depending on the number of partitions available in the EventHub instance. I've updated the integration tests with an EH with 10 partition, in order to reproduce the same routing anyone testing should use 10 partitions as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ az eventhubs namespace create --name debezium-test --resource-group eventhubstes
 
 ### Create an Event Hub
 
-Create an Event Hub (equivalent to a topic) with `one` partition. Check the documentation for options on how do this using the [Azure Portal](https://docs.microsoft.com/azure/event-hubs/event-hubs-create#create-an-event-hub), [Azure CLI](https://docs.microsoft.com/azure/event-hubs/event-hubs-quickstart-cli#create-an-event-hub) etc. , e.g. on the CLI:
+Create an Event Hub (equivalent to a topic) with 10 partitions. Check the documentation for options on how do this using the [Azure Portal](https://docs.microsoft.com/azure/event-hubs/event-hubs-create#create-an-event-hub), [Azure CLI](https://docs.microsoft.com/azure/event-hubs/event-hubs-quickstart-cli#create-an-event-hub) etc. , e.g. on the CLI:
 
 ```shell
-az eventhubs eventhub create --name debezium-test-hub --resource-group eventhubstest --namespace-name debezium-test
+`az eventhubs eventhub create` --name debezium-test-hub --resource-group eventhubstest --namespace-name debezium-test --partition-count 10
 ```
 
 ### Build the module
--- a/debezium-server-eventhubs/pom.xml
+++ b/debezium-server-eventhubs/pom.xml
@@ -38,10 +38,11 @@
             <artifactId>debezium-testing-testcontainers</artifactId>
             <scope>test</scope>
         </dependency>
-<dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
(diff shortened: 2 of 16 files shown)


jpechane:
@slknijnenburg Hi, thanks for the PR. Please take a look at https://debezium.io/documentation/reference/2.3/transformations/partition-routing.html
I wonder if we can delegate parttion computation to it only and just use the partiiton number provided by it just spread over the partition count defined in EventHubs?

slknijnenburg:
@jpechane Yes that makes a lot of sense! Still requires the batch management changes, but I can clean up the hashing/routing part. Pretty weird I missed that SMT when looking into options, thanks for the tip. I'll mark the PR as ready to review again once the changes are implemented.

slknijnenburg:
@jpechane I spent some time digging into the PartitionRouter, and it should do exactly what's needed to make the EventHubs partitioning support work as well. 

However, I'm stuck on the following: the partition info is not available in the ChangeEvent<> that's passed into the consumer's handleBatch() method. I _think_ it should be available in the EmbeddedEngineChangeEvent's `sourceRecord` property, but that implementation is private and cannot be accessed in the EventHubsChangeConsumer. Also, when running an integration test for the EventHubsChangeConsumer that has a PartitionRouterSMT configured I don't see any change in the sourceRecord.sourcePartition property, which makes me doubt the SMT is functioning or whether it's result reaches the EmbeddedEngineChangeEvent.

Can you point me in a direction of where you expect the partition property to be accessible in the EventHubsChangeConsumer, or what changes would be needed to make it accessible?

jpechane:
@slknijnenburg I think it make sense to expose the value in the interface.

brenoavm:
Hi, @slknijnenburg, do you have any updates about this?

slknijnenburg:
> Hi, @slknijnenburg, do you have any updates about this?

@brenoavm Besides a wonderful long holiday I had to focus on some other priorities. I intend to get back to this PR by the end of next week.

slknijnenburg:
I'm closing this PR in favor of https://github.com/debezium/debezium-server/pull/51 that contains an implementation leveraging the PartitionRouting SMT.

PR #35, opened 2023-08-07
DBZ-6718 PoC

MartinMedek:


MartinMedek:
I guess the main class to thing at is the TestContainerPrototype, which contains test run on docker image and on a jar artifact. It seems that running tests on native executable is dead end, since debezium is not buildable this way.

PR #40, opened 2023-09-14
DBZ-6909 Kinesis sink sending a partition key bigger than 256 bytes

brenoavm:


jpechane:
@brenoavm Alternative solution proposed in https://issues.redhat.com/browse/DBZ-6909?focusedId=23030371&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23030371

jpechane:
In fact a similar fetautre would be handy for https://github.com/debezium/debezium-server/pull/33#issuecomment-1662102270 as well

RogerYeh1:
is this still in progress? kinesis now needs the legnth of partitionKey to be less than or equals to 256.
i've tried the above solutions(partitionRouting SMT) it is still now working for me though.

jpechane:
@RogerYeh1 Hi, it seems the PR has stalled. Would you be willing to take it over and finish it? Thanks

brenoavm:
@Naros I tried to rebase my branch and sign off the last commit but it is still getting check errors.

There are other commits on the log without the sign, is that correct? How can I handle that?

Naros:
It's not an issue on your end @brenoavm, it's the fact the GHA wasn't able to delete a comment left by my user account, and that is what caused the issue. It should be resolved now.

PR #43, opened 2023-10-13
DBZ-7032 Implement Retries for Amazon Kinesis Event Transmittion

ilyasahsan123:
https://issues.redhat.com/browse/DBZ-7032

### DIFF
--- a/debezium-server-kinesis/src/main/java/io/debezium/server/kinesis/KinesisChangeConsumer.java
+++ b/debezium-server-kinesis/src/main/java/io/debezium/server/kinesis/KinesisChangeConsumer.java
@@ -11,7 +11,6 @@ import java.util.List;
 import java.util.Optional;
 
 import io.debezium.DebeziumException;
-import io.debezium.annotation.VisibleForTesting;
 import io.debezium.util.Clock;
 import io.debezium.util.Metronome;
 import jakarta.annotation.PostConstruct;
@@ -61,11 +60,8 @@ public class KinesisChangeConsumer extends BaseChangeConsumer implements Debeziu
     private String region;
     private Optional<String> endpointOverride;
     private Optional<String> credentialsProfile;
-
-    private static int retries;
-    private static Duration retryInterval;
     private static final int DEFAULT_RETRIES = 5;
-    private static final Long RETRY_INTERVAL = Integer.toUnsignedLong(1_000); // Default to 1s
+    private static final Long RETRY_INTERVAL = Integer.toUnsignedLong(1_000);
 
     @ConfigProperty(name = PROP_PREFIX + "null.key", defaultValue = "default")
     String nullKey;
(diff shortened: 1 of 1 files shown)


debezium-server-kinesis/src/main/java/io/debezium/server/kinesis/KinesisChangeConsumer.java
```diff
@@ -101,6 +118,23 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
             throws InterruptedException {
         for (ChangeEvent<Object, Object> record : records) {
             LOGGER.trace("Received event '{}'", record);
+
+            int attempts = 0;
+            if (!recordSent(record)) {
```
vjuranek, line 123: you need some while loop to retry sending the records when it fails, see e.g. https://github.com/debezium/debezium-server/blob/main/debezium-server-http/src/main/java/io/debezium/server/http/HttpChangeConsumer.java#L164

vjuranek (approved):
thanks, LGTM now @ilyasahsan123 !

debezium-server-kinesis/src/main/java/io/debezium/server/kinesis/KinesisChangeConsumer.java
```diff
@@ -55,6 +60,8 @@ public class KinesisChangeConsumer extends BaseChangeConsumer implements Debeziu
     private String region;
     private Optional<String> endpointOverride;
     private Optional<String> credentialsProfile;
+    private static final int DEFAULT_RETRIES = 5;
+    private static final Long RETRY_INTERVAL = Integer.toUnsignedLong(1_000);
```
brenoavm, line 65: It would be interesting if the DEFAULT_RETRIES and RETRY_INTERVAL were config properties. Something like debezium.sink.kinesis.default.retries and debezium.sink.kinesis.retry.interval. WDYT?

ilyasahsan123: [CMIIW] I am not sure about this; it's technically possible, but in my opinion, it's not necessary. If we take a look at other sinks:

1. For HTTP, the retries are hardcoded to 5.
2. For EventHub, the client uses the default retries, which is 9 times.

The exception is Pub/Sub, where the retry properties can be dynamically set via properties.

Do you have any comment @vjuranek?

vjuranek:
@ilyasahsan123 the CI fails due to wrong formatting. Could you please take a look and fix it (I guess something like `mvn -f server/debezium-server-kinesis net.revelc.code.formatter:formatter-maven-plugin:2.20.0:format` should do the job). Thanks!

ilyasahsan123:
> @ilyasahsan123 the CI fails due to wrong formatting. Could you please take a look and fix it (I guess something like `mvn -f server/debezium-server-kinesis net.revelc.code.formatter:formatter-maven-plugin:2.20.0:format` should do the job). Thanks!

hi @vjuranek, thanks for the suggestion. I have formatted the code. Could you please take a look? thanks

PR #44, opened 2023-10-25
DBZ-7077 Use jakarta.inject.Inject instead of com.google.inject.Inject

vjuranek:
https://issues.redhat.com/browse/DBZ-7077
https://issues.redhat.com/browse/DBZ-7092


vjuranek:
There seems to be more issues with HTTP tests.

jpechane:
@vjuranek Applied thanks

PR #49, opened 2023-11-22
Feature Request: DBZ-6703 Support native RabbitMQ Streams

ahmedrachid:
Related feature request: https://issues.redhat.com/browse/DBZ-6703

**Description:** 
The goal of this PR is to add and support native RabbitMQ Streams in the Debezium server. It includes a new integration module leveraging RabbitMQ Streams Java API, new configuration called "rabbitmqstream" and tests.

**Value Added:**
- Support native RabbitMQ Streams, for high throughput, real-time data streaming to RabbitMQ


jpechane (changes requested):
@ahmedrachid Thanks for the PR. Unfortunately Pubsub/PubSubLite testing is not working as intended so your PR will need to take a bit different approach to it.
I left couple of comments on what has to be changed to get everything running.
The key thing you need to remeber is using the system properties `-Ddebezium.sink.type=rabbitmqstream -Dtest.type=IT` when running `RabbitMqStreamIT` and the need to change the configuration in a way I commented on test config source.
That way you should get it running correctly for the new implementation and test.
When done we'll discuss what is necessary to be done in `pom.xml` to make sure that with Maven build both the old and new tests are executed and properly configured.

ahmedrachid:
Hello @jpechane , changes/fix for issues are made: i'm running my test in this way: 

-ea -Ddebezium.sink.type=rabbitmqstream -Dtest.type=IT -Ddebezium.sink.rabbitmqstream.stream=testc.inventory.customers

Could you please review ?

jpechane:
@ahmedrachid I did few changes and made sure the tests are executted as part of the build. PR applied, thanks a lot!

ahmedrachid:
Great news ! thanks @jpechane

PR #55, opened 2023-12-09
DBZ-6071 Add support for streamNameMapper in Kafka sink

PlugaruT:


PlugaruT:
The only thing is that I don't know how to limit the injection of a custom mapper to  the scope of a single test. Open to suggestions.

PlugaruT:
CI is failing not related to the change

PlugaruT:
> @PlugaruT Applied, thanks. Could you please send a PR for docs too? IMHO we don't test custom mappers per sink so it is fine.

Sure, will do later today.

PR #56, opened 2023-12-11
DBZ-7216 Adjusting to changed CE type value

gunnarmorling:
@jpechane, @Naros, looks like a missing leftover from DBZ-7216 to me. That being said, I'm somewhat doubtful about this change in the first place:

* It seems unexpected in a non-minor release?
* If it actually should be changed, the [documentation](https://debezium.io/documentation/reference/stable/integrations/cloudevents.html) needs updating as well
* Is there a CI build missing which would have flagged this early on?

### DIFF
--- a/debezium-server-http/src/test/java/io/debezium/server/http/HttpIT.java
+++ b/debezium-server-http/src/test/java/io/debezium/server/http/HttpIT.java
@@ -150,12 +150,12 @@ public class HttpIT {
                 };
                 hm = om.readValue(request.getBody(), tref);
 
-                Assertions.assertEquals("/debezium/postgresql/testc", (String) hm.get("source"));
-                Assertions.assertEquals("io.debezium.postgresql.datachangeevent", (String) hm.get("type"));
-                Assertions.assertEquals("1.0", (String) hm.get("specversion"));
-                Assertions.assertEquals("postgres", (String) hm.get("iodebeziumdb"));
-                Assertions.assertEquals("inventory", (String) hm.get("iodebeziumschema"));
-                Assertions.assertEquals("customers", (String) hm.get("iodebeziumtable"));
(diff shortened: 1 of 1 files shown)


jpechane:
@gunnarmorling Hmm, thanks for pointing this out. I missed the inmpact on backward compatibility with this one.
In fact there is a follow-up https://github.com/debezium/debezium/pull/5083 where I argue exactly the same issue

@rkudryashov It seems there is another problem related to the schema naming and backward compatibility.
Maybe we should rethink the pattern-based approach I proposed so there will be backward compatible config and you could do arbitrary configuration?

rkudryashov:
@jpechane the mentioned PR consists of two parts:
1) ability to customize CE envelope schema name
2) ability to customize CE.data schema name

In both cases, I haven't changed their default values unlike CE.type name in this issue. So #5083 should not introduce a similar problem and it is already backward compatible. The only problem is a configuration option for 2). Need to think a little more how not to make it a breaking change requiring the user to set `dataSchemaName` - but not sure how to do it

What do you mean by pattern-based approach? Is it somehow related to [this](https://github.com/debezium/debezium/blob/5014e2ace3e712a819cbae3fb25dc6be5aff517c/debezium-core/src/main/java/io/debezium/converters/spi/CloudEventsMaker.java#L170) pattern: server name + database name + schema name?

jpechane:
@rkudryashov Applied, thanks

@gunnarmorling You are right this is breaking change but upon further consideration I believe this is justifiable.
Really if you look in the code the schema naming was inconsistent, using `connector` in it in some cases and not using it in others. Now it is all the same. WRT the change of lettercase. Again in the other case it is `Data` with upper case. So it makes sense to align it for consistency PoV. And as the change in schema prefix is already breaking change then it is good to do it in a single step.

PR #57, opened 2023-12-11
DBZ-7244 Event Hubs topic routing

aleksandervalle:
Fixes https://github.com/debezium/dbz/issues/1173

https://issues.redhat.com/browse/DBZ-7244

### DIFF
--- a/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
+++ b/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
@@ -48,15 +48,13 @@ public class EventHubsChangeConsumer extends BaseChangeConsumer
     private static final String PROP_PREFIX = "debezium.sink.eventhubs.";
     private static final String PROP_CONNECTION_STRING_NAME = PROP_PREFIX + "connectionstring";
     private static final String PROP_EVENTHUB_NAME = PROP_PREFIX + "hubname";
-    private static final String PROP_EVENTHUB_NAME_LIST = PROP_PREFIX + "hubname.list";
     private static final String PROP_PARTITION_ID = PROP_PREFIX + "partitionid";
     private static final String PROP_PARTITION_KEY = PROP_PREFIX + "partitionkey";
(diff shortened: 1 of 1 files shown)


aleksandervalle:
Thinking about adding an integration test, but would have to delve deeper into how tests are setup first, and how to add changes from another source table (to test topic routing for two different tables).

Seems like the [second commit](https://github.com/debezium/debezium-server/pull/57/commits/a24d1809ce08d586018881db0be7c054179ae36a) potentially fixes an earlier regression. I'm not sure how I would configure a `customProducer`, but it doesn't seem like a `customProducer` would work (as it's not wrapped in a `BatchManager`).

jpechane:
@aleksandervalle Overall LGTM, thanks. I've one proposal, instead of adding a new config option I think it is ok just to modify the existing one to be able to process the list instead of a single value. It will be backward compatible, the code will be simplified and no new option is needed.

aleksandervalle:
> @aleksandervalle Overall LGTM, thanks. I've one proposal, instead of adding a new config option I think it is ok just to modify the existing one to be able to process the list instead of a single value. It will be backward compatible, the code will be simplified and no new option is needed.

I removed the new optional config, and we're now using the required `debezium.sink.eventhubs.hubname` config. Using `eventHubNames[0]` as fallback in case `record.destination()` does not match any of the given event hub names in the configuration. This should ensure backwards compatibility, as far as I can see.

aleksandervalle:
Maybe the [documentation](https://debezium.io/documentation/reference/stable/operations/debezium-server.html#hub-name) should be updated to reflect these changes? Idea for new description:

Specifies the name(s) of the Event Hub(s) to be used. You can provide a single hub name or a comma-separated list of hub names without any whitespace. When multiple hub names are specified, Debezium will direct the change events to the corresponding Event Hub based on the event's destination. If the destination of a change event does not match any of the listed hub names, the first hub in the list will be used as the default destination.

**Usage Example**: `debezium.sink.eventhubs.hubname=hub1,hub2,hub3`

In this example, `hub1` serves as the default Event Hub if no specific matching is found for a change event.

jpechane:
@Jokero Would you be able to help with getting it finished?

PR #61, opened 2024-01-16
DBZ-7117 Prevent OOM in Debezium Redis sink

eizners:


debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -26,55 +25,79 @@ public class RedisMemoryThreshold {
...
+     * @return
+     */
+    public boolean checkMemory(long extraMemory, int bufferSize, int bufferFillRate) {
+        Tuple2<Long, Long> memoryTuple = memoryTuple();
+
+        maximumMemory = memoryTuple.getItem2();
+
+        if (maximumMemory == 0) {
+            totalProcessed += bufferSize;
+            LOGGER.debug("Total Processed Records: {}", totalProcessed);
```
Naros, line 64: If `maximumMemory` is set to `0`, I assume we are not doing any memory check, and is there a case where `totalProcessed` here would just continue to grow until it overflows potentially?  If so, I wonder if it would be better to avoid incrementing and simply just return `true` only?

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -26,55 +25,79 @@ public class RedisMemoryThreshold {
...
+            accumulatedMemory = prevAccumulatedMemory;
+            return false;
+        }
+        else {
+            LOGGER.debug(
+                    "Maximum reached: {}; Used Mem {}; Max Mem: {}; Accumulate Mem: {}; Estimated Used Mem: {}, Record Size: {}, NumRecInBuff: {}; Total Processed Records: {}",
+                    (estimatedUsedMemory >= maximumMemory), humanReadableSize(usedMemory), humanReadableSize(maximumMemory),
+                    humanReadableSize(accumulatedMemory), humanReadableSize(estimatedUsedMemory),
+                    humanReadableSize(extraMemory), bufferSize, totalProcessed + bufferSize);
+            totalProcessed += bufferSize;
```
Naros, line 95: Same as above, could this eventually overflow?

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -26,55 +25,79 @@ public class RedisMemoryThreshold {
...
-            if (percentage >= memoryThreshold) {
-                LOGGER.info("Sink memory threshold percentage was reached. Will retry; " +
-                        "(current: {}%, configured: {}%, used_memory: {}, maxmemory: {}).",
-                        percentage, memoryThreshold,
-                        usedMemory, maxMemory);
-                return false;
-            }
+
+        long extimatedBatchSize = extraMemory * bufferFillRate;
+        long usedMemory = memoryTuple.getItem1();
```
Naros, line 69: In the `memoryTuple()` function, the first element stored here could technically be `null` if the key for this element is not found and could result in a `NullPointerException`.  Similar to how you guarantee there is a non-null value for item2, you should do the same for item1.

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -91,43 +114,55 @@ private Tuple2<Long, Long> memoryTuple(long defaultMaxMemory) {
...
+            return "0 B";
+        }
+
+        while (sizeInUnit >= 1024 && unitIndex < units.length - 1) {
+            sizeInUnit /= 1024.0;
+            unitIndex++;
+        }
+
+        return String.format("%.2f %s", sizeInUnit, units[unitIndex]);
+    }
```
Naros, line 167: I'd suggest creating a follow-up Jira and consider moving this to `debezium-core` inside either the `io.debezium.util.NumberConversions` class or `io.debezium.util.Strings` utility classes as I think this could be helpful for any connector that may want to log similar bits in the future, perhaps with a variant that uses the `"%.2f %s" format by default and another that takes a format string that allows customization.

Naros (commented):
A few inline comments @eizners.

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
spicy-sauce: I'd suggest to name this with a verb, something like:
- formatReadableSize
- getSizeInHumanReadableFormat
- convertSizeToReadableString
- getFormattedSizeString

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
eizners: This is correct is maximumMemory we are not doing any checks. The totalProcessed is long so I dobt that Debezium will ever process 9,223,372,036,854,775,807 records without restart

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
eizners: Same answer as above

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
eizners: Fixed

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
eizners: Changing to getSizeInHumanReadableFormat. I will create a new Jira for that.  This fix will need to go into different repository

jpechane:
@eizners Applied vi https://github.com/debezium/debezium-server/pull/63, thanks a lot!

PR #64, opened 2024-01-25
DBZ-7417 Kinesis Sink Exception on PutRecord

brenoavm:
https://issues.redhat.com/browse/DBZ-7417


jpechane:
@brenoavm Applied, thanks!

PR #65, opened 2024-01-25
DBZ-6955 Treat HTTP/2 GOAWAY responses as retriable

Naros:
https://issues.redhat.com/browse/DBZ-6955


jpechane:
@Naros Applied, thanks!

PR #66, opened 2024-02-02
feat: added an option to set a custom unique ordering key

ArthurLERAY:
https://issues.redhat.com/browse/DBZ-7435


jpechane:
@ArthurLERAY I've cleaned the PR and made the property optional. Could you please prepare a PR a docs PR describing the new property?
Thanks

PR #68, opened 2024-02-10
DBZ-7470 Add option to specify embedded engine factory builder

vjuranek:
Add option to specify embedded engine factory builder and switch HTTP tests to async embedded engine.

https://issues.redhat.com/browse/DBZ-7470


jpechane:
@vjuranek Applied, thanks!

PR #70, opened 2024-02-19
DBZ-7515 Run tests against Apicurion in a single thread

vjuranek:
https://issues.redhat.com/browse/DBZ-7515


jpechane:
@vjuranek Applied, thanks!
Test failure is the RocketMq server refusing to accept messages if free disk space falls below a certain threshold.

PR #73, opened 2024-02-21
DBZ-7545 Add second dependencySet for implicit jakarta inclusion

obabec:
https://issues.redhat.com/browse/DBZ-7545


jpechane:
@obabec Applied, thanks

PR #75, opened 2024-02-26
DBZ-7564 Wait for Redis server to start

vjuranek:
https://issues.redhat.com/browse/DBZ-7564


jpechane:
@vjuranek Applied, thanks!

PR #76, opened 2024-02-28
DBZ-7578 Use short flush interval in Redis offset tests

vjuranek:
https://issues.redhat.com/browse/DBZ-7578


jpechane:
@vjuranek Applied, thanks

PR #78, opened 2024-03-01
DBZ-7585 Don't miss event when current batch is full

vjuranek:
https://issues.redhat.com/browse/DBZ-7585


jpechane:
@vjuranek Applied, thanks! We never got a word from the reporter whom we asked to test the change. For that reason I am going  not to backport the fix into the 2.5 so provide a fallback path in case of any issue.

PR #79, opened 2024-03-01
DBZ-7576 Use fixed port for Redis server

vjuranek:
When the Redis container uses random port, even single test fails with `Unexpected end of stream` exception on my machine. I observed that it starts working when Redis uses fixed port. Unfortunatelly I have no clue what is the root cause.

Try to switch to fixed port and see if it helps also on the CI and the testsuite will be more stable. Or vice versa - if it makes things worse and there will be port conflicts.

https://issues.redhat.com/browse/DBZ-7576


jpechane:
@vjuranek Applied, thanks, let's hope it solves the issue!

PR #83, opened 2024-03-21
Fix RedisSchemaHistoryIT tests

vjuranek:
https://issues.redhat.com/browse/DBZ-7692
https://issues.redhat.com/browse/DBZ-7701


vjuranek:
Still may fail due to [DBZ-7701](https://issues.redhat.com/browse/DBZ-7701), which is another issue (and here have no clue what the root cause could be).

jpechane:
@vjuranek Applied, thanks!

PR #86, opened 2024-04-29
DBZ-7829 Added nats JWT/seed authentication config options

jpechane:
Added two new properties; `auth.jwt` and `auth.seed` to allow to pass a JWT and a seed to connect to nats when it has been configured for decentralized auth (like Synadia)


calmera:
Hmm, I see Quarkus is considering the new options as being mandatory (default=""), which probably means it cannot be an empty string. I updated my branch so they use a "-" instead as the default value and updated the condition for checking whether or not the fields should be taken into account

jpechane:
@calmera I prepared https://github.com/debezium/debezium-server/pull/87 I just forget to push it

PR #89, opened 2024-05-09
DBZ-7847 Fix apicurio deps alignment

Naros:
Needed for changes in core related to https://github.com/debezium/debezium/pull/5544


jpechane:
@Naros Applied, thanks

PR #94, opened 2024-05-27
Fix for an exception in PubSubChangeConsumer

ankurg03:


jpechane:
LGTM, thanks. Could you please create a Jira issue for the change?

jpechane:
I see this change is also part of https://github.com/debezium/debezium-server/pull/95
So maybe this one can be closed?

PR #95, opened 2024-05-28
DBZ-7904 Forced shutdown after failed controlled shutdown on errored Pub/Sub managed channel

ankurg03:


jpechane:
@ankurg03 Could you please create a Jira issue for the change describing the problem and fix?

debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
```diff
@@ -213,12 +214,33 @@ void close() {
...
+                LOGGER.warn("Exception while closing publisher: {}", e.getMessage(), e);
             }
         });
+        shutdownChannel(channel);
+    }
 
+    void shutdownChannel(ManagedChannel channel) {
         if (channel != null && !channel.isShutdown()) {
-            channel.shutdown();
+            ExecutorService executor = Executors.newSingleThreadExecutor();
```
jpechane, line 225: Please use `io.debezium.util.Threads` class

debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
```diff
@@ -213,12 +214,33 @@ void close() {
...
+                channel.shutdown();
+                try {
+                    channel.awaitTermination(channelShutdownTimeout, TimeUnit.SECONDS);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+            });
+
+            try {
+                future.get(channelShutdownTimeout + 1, TimeUnit.SECONDS);
```
jpechane, line 236: Timeout is in milliseconds but `get` is in seconds. Also why +1?

There should be also `0` as a special value that will do shutdown as is done currently.

debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
```diff
@@ -213,12 +214,33 @@ void close() {
...
+            } catch (TimeoutException e) {
```
jpechane, line 237: I think it is fine to use just generic `catch (Exception e)`

jpechane (commented):
@ankurg03 Thanks for the PR. I left couple of comments.

ankurg03:
> @ankurg03 Could you please create a Jira issue for the change describing the problem and fix?

https://issues.redhat.com/browse/DBZ-7904

debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
ankurg03: The io.debezium.util.Threads class is designed to work with SourceConnector, but I want to utilize it within a ChangeConsumer context.

jpechane: @ankurg03 Right, could you please then try to create the executor in the way that the name of the executor thread is more desciptive in the sense that it is clear that it is

* from Debezium
* to which sink it relates
* what is its purpose

Maybe you can even create a limited equivalent of `Threads` class in Debezim Server core module so we can re-use it elswhere if possible.

ankurg03: Is there a reason for ThreadFactory to extend SourceConnector in the Threads class? The current syntax is as follows:

ThreadFactory threadFactory(Class<? extends SourceConnector> connector, String connectorId, ....)
Would it be better to change the syntax to something like this:

public static ThreadFactory threadFactory(Class<?> connectorType, String connectorId, ....)
This way, the same class can be used for both sink and source processors.

jpechane: It was intentinal as we expected that `SourceConnector` class might provide some common metadata in future we'd rely on. In the hindsight it is probably unnecessary so yet, it could be converted to generic class. In that case I'd also rename argument from `connector` to `component` to reflect its more generic state.

ankurg03: Sure, I will raise a pull request to implement this change. i.e. I will make the ThreadFactory in the Threads class more generic.

ankurg03: I've updated the PR. Could you please take a look?

jpechane:
@ankurg03 Thanks, I reorganized and for-pushed commits. The PR will be merged when CI passes.

jpechane:
@ankurg03 Applied, thanks!

PR #97, opened 2024-06-02
DBZ-7913 Added Prometheus JMX exporter to server distribution

jcechace:
https://issues.redhat.com/browse/DBZ-7913


jpechane:
@jcechace Applied, thanks

PR #98, opened 2024-06-04
DBZ-7922: Add support of TLS auth to NATS JetStream sink

mthjs:


jpechane:
@mthjs Applied, thanks

jpechane:
@mthjs Could you please create a docs PR for the new options? Thanks

PR #99, opened 2024-06-11
DBZ-7767 Allow empty config property values

Naros:
https://issues.redhat.com/browse/DBZ-7767


jpechane:
@Naros Applied, thanks

PR #105, opened 2024-07-08
DBZ-8036 Added `type` to Prometheus JMX exporter

missingcharacter:
Otherwise metrics show up as `untyped`
Also added `kafka.producer` metrics since this is not going to be used with a `kafka.connect` container

### DIFF
--- a/debezium-server-dist/src/main/resources/distro/conf/metrics.yml
+++ b/debezium-server-dist/src/main/resources/distro/conf/metrics.yml
@@ -3,14 +3,17 @@ ssl: false
 lowercaseOutputName: false
 lowercaseOutputLabelNames: false
 rules:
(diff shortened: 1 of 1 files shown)


jpechane:
@missingcharacter Thanks for the PR. Could you please create a Jira describing the change and change the commit message to something like `DBZ-nnnn Added type to Prometheus JMX exporter`. Also I see you do two unrelated changes so maybe it would be better to split the Jira into two and the current commit into two for those two changes.

missingcharacter:
@jpechane I created https://issues.redhat.com/browse/DBZ-8036 for this Pull Request

missingcharacter:
@jpechane I made a separate PR for the kafka producer metrics https://github.com/debezium/debezium-server/pull/106

jpechane:
@missingcharacter LGTM, applied, thanks!

PR #111, opened 2024-08-02
DBZ-8108 implemente option AddressResolver for rabbitmq native stream

LucasZhanye:
When use rabbitmq native stream as sink for dbz, it will produce error as like: "Error while creating stream connection to fdb49ba7df14:5552. fdb49ba7df14: Name or service not known. This may be due to the usage of a load balancer that makes topology discovery fail. Use a custom AddressResolver or the --load-balancer flag if using StreamPerfTest".

From the docs https://rabbitmq.github.io/rabbitmq-stream-java-client/stable/htmlsingle/#understanding-connection-logic
We need to use options "AddressResolver" to let it connect successful


jpechane:
@LucasZhanye Applied, thanks!

PR #113, opened 2024-08-05
DBZ-8118 Rabbitmq native stream support individual stream for each table

LucasZhanye:
This is a new feature for rabbitmq native stream.

Currently, all table changes are send to one stream.I think it is not very well for many situation.

So I change it, and add some options for it and it doesn't affect current user.

### DIFF
--- a/debezium-server-bom/pom.xml
+++ b/debezium-server-bom/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>io.debezium</groupId>
         <artifactId>debezium-build-parent</artifactId>
-        <version>3.0.0.Alpha2</version>
+        <version>3.0.0-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>debezium-server-bom</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>io.debezium</groupId>
         <artifactId>debezium-parent</artifactId>
-        <version>3.0.0.Alpha2</version>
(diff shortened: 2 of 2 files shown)


jpechane:
@LucasZhanye Hi, thanks for the PR. It looks really good! I've only two comments

1) As I look in the code I belive that maybe the approach of having Map or static attribute is unnecessay optimization. I propose to handle both cases in the same way, so the map will always be used just in one case the key will be the static value, in other the re-mapped topic name. It will definitely simplify the code.
2) I propse to modify the existing integration test and check if the test message is coming from the expected stream

LucasZhanye:
> @LucasZhanye Hi, thanks for the PR. It looks really good! I've only two comments
> 
> 1. As I look in the code I belive that maybe the approach of having Map or static attribute is unnecessay optimization. I propose to handle both cases in the same way, so the map will always be used just in one case the key will be the static value, in other the re-mapped topic name. It will definitely simplify the code.
> 2. I propse to modify the existing integration test and check if the test message is coming from the expected stream

@jpechane hi，thanks for your advice. 
1. What you mean is that I should only use Map to cache all condition(whether single stream or not)? and I am not clearly with that streamMaxAge/streamMaxLength/streamMaxSegmentSize properties is neccessary?
2. The integration test what you said, I don't know how to do it, can you give me a example? The source code location or other?

jpechane:
Please see `io.debezium.server.rabbitmq.RabbitMqIT`
For the former question, just imagine that you'll remove `producer` attribute completely and use `streamProducers` only. Also in no circumstance the `connect` method will contain creation of producer. Is it a bit clearer?

PR #114, opened 2024-08-06
Rabbitmq native streams support separate streams per table and fix st…

LucasZhanye:
This is a new feature for rabbitmq native stream.

Currently, all table changes are send to one stream.I think it is not very well for many situation.

So I change it, and add some options for it and it doesn't affect current user.

And I fix that without specifying username and password and virtual host when connect,it will use guest/guest rather than setting of properties.


jpechane:
@LucasZhanye Thanks a lot! Merged via https://github.com/debezium/debezium-server/pull/115

PR #116, opened 2024-08-13
DBZ-7575 Resolve all futures before marking a batch as completed in Kafka sink

PlugaruT:
Resolve futures before marking a batch as completed for Kafka sink

This will make sure that records are being send or we crash the sink instead of sink being blocked forever.

### DIFF
--- a/debezium-server-kafka/src/main/java/io/debezium/server/kafka/KafkaChangeConsumer.java
+++ b/debezium-server-kafka/src/main/java/io/debezium/server/kafka/KafkaChangeConsumer.java
@@ -10,6 +10,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
@@ -25,6 +27,7 @@ import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,7 +49,11 @@ public class KafkaChangeConsumer extends BaseChangeConsumer implements DebeziumE
     private static final Logger LOGGER = LoggerFactory.getLogger(KafkaChangeConsumer.class);
 
     private static final String PROP_PREFIX = "debezium.sink.kafka.";
+
     private static final String PROP_PREFIX_PRODUCER = PROP_PREFIX + "producer.";
+
+    @ConfigProperty(name = PROP_PREFIX + "wait.message.delivery.timeout.ms", defaultValue = "30000")
+    Integer waitMessageDeliveryTimeout;
 
     private KafkaProducer<Object, Object> producer;
 
@@ -91,15 +98,16 @@ public class KafkaChangeConsumer extends BaseChangeConsumer implements DebeziumE
                 Headers headers = convertKafkaHeaders(record);
 
                 String topicName = streamNameMapper.map(record.destination());
-                Future<RecordMetadata> recordMetadataFuture = producer.send(new ProducerRecord<>(topicName, null, null, record.key(), record.value(), headers), (metadata, exception) -> {
-                    if (exception != null) {
-                        LOGGER.error("Failed to send record to {}:", topicName, exception);
-                        throw new DebeziumException(exception);
-                    }
-                    else {
-                        LOGGER.trace("Sent message with offset: {}", metadata.offset());
-                    }
-                });
+                Future<RecordMetadata> recordMetadataFuture = producer.send(new ProducerRecord<>(topicName, null, null, record.key(), record.value(), headers),
+                        (metadata, exception) -> {
+                            if (exception != null) {
+                                LOGGER.error("Failed to send record to {}:", topicName, exception);
+                                throw new DebeziumException(exception);
+                            }
+                            else {
+                                LOGGER.trace("Sent message with offset: {}", metadata.offset());
+                            }
+                        });
                 futures.add(recordMetadataFuture);
                 committer.markProcessed(record);
             }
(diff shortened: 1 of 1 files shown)


jpechane:
Please take a look at `io.debezium.server.pubsub.PubSubChangeConsumer.handleBatch(List<ChangeEvent<Object, Object>>, RecordCommitter<ChangeEvent<Object, Object>>)`. I believe we should add the timeout to the `Future.get()` here too.

PlugaruT:
@jpechane if changes look good, any chance this PR get's back-ported to 2.x versions?

jpechane:
@PlugaruT LGTM. I am definitely fine with backporting it. Just please do one more change. If the timeout is set to `0` then use `get()` call so the timeout can be disabled.

debezium-server-kafka/src/main/java/io/debezium/server/kafka/KafkaChangeConsumer.java
```diff
@@ -81,31 +91,44 @@ void stop() {
...
+                            if (exception != null) {
+                                LOGGER.error("Failed to send record to {}:", topicName, exception);
+                                throw new DebeziumException(exception);
+                            }
+                            else {
+                                LOGGER.trace("Sent message with offset: {}", metadata.offset());
+                            }
+                        });
+                futures.add(recordMetadataFuture);
                 committer.markProcessed(record);
```
jpechane, line 112: @PlugaruT I am sorry I completely forgot to point you to https://issues.redhat.com/browse/DBZ-7575?focusedId=25342060&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25342060

Generally this markProcessed needs to be removed here and moved to the following loop that processes the futures.
And this brings one more issue - when futures are used, are the records delivered in sequence? If not I see two problems
1) Two records for the same PK can be delivered in an incorrect order
2) Theprocessing mark is not done in correct order which I belive can cause issue with offsets

PlugaruT: I actually have been thinking about this, but then got confused as you said one thing, and Vojtech another 😅 No probs with that. 

Reading the `producer.send` [docs](https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send-org.apache.kafka.clients.producer.ProducerRecord-), the way we use it here, with a callback, seems that it's a non-blocking call and from my understanding, the batching and ordering is done by the producer itself since it has some buffering logic. 

Now that I think about it, maybe it's worth actually reverting this change and going back to just add `latch.countDown();` in the callback itself and that would be it. 
Let me know what you think.

jpechane: In fact, is not the real issue that the code should be like
```
                producer.send(new ProducerRecord<>(topicName, null, null, record.key(), record.value(), headers), (metadata, exception) -> {
                    if (exception != null) {
                        LOGGER.error("Failed to send record to {}:", topicName, exception);
                        throw new DebeziumException(exception);
                    }
                    else {
                        LOGGER.trace("Sent message with offset: {}", metadata.offset());
                        latch.countDown();
                    }
                }).get();
```
Note the `get()` at the end. In this case if the exception is thrown it is passed up to the call and the Server will terminate. In this case it still makes sense so keep the new option, just either call get, or get with timeout. WDYT?

PlugaruT: Yes, makes sense to me. 
I applied the changes. Let me know if it works. Thanks

jpechane:
@PlugaruT Applied, thanks! Cloud you please provide a PR with docs update too?

PlugaruT:
> @PlugaruT Applied, thanks! Cloud you please provide a PR with docs update too?

I will have it by end of week!

debezium-server-kafka/src/main/java/io/debezium/server/kafka/KafkaChangeConsumer.java
```diff
@@ -81,31 +88,35 @@ void stop() {
...
+                            }
+                            else {
+                                LOGGER.trace("Sent message with offset: {}", metadata.offset());
+                            }
+                        });
+                if (waitMessageDeliveryTimeout == 0) {
+                    recordMetadataFuture.get();
+                }
+                else {
+                    recordMetadataFuture.get(waitMessageDeliveryTimeout, TimeUnit.MILLISECONDS);
```
rk3rn3r, line 111: Why do we have our own timeout handling here instead of using `delivery.timeout.ms` option for the KafkaProducer / ProducerConfig?
Are you aware that the KafkaProducer has this config option and it is set to 120_000 ms = 2 minutes?
You default to 30_000 / 30 seconds. That can lead to more errors than they really happen because you only wait for 30 seconds while the producer might return after 2 mins, when I understand the code correctly?!
Am I wrong here? @jpechane ?

jpechane: This is good point. The code still should stay in place as a defensive measure but the default should be changed to correspond to Kafka default (or slightly more).

PlugaruT: I can open a PR really quick and set the default value to match the `delivery.timeout.ms` config.
What I've noticed in our case is that even with this timeout, the producer might retry indefinitely to push some records to Kafka since default values are very permisive and we ended up with WAL increasing but Debezium not crashing. I feel like this config will allow us to crash hard the server and rely on a full restart for a recovery, maybe...

RichardFourie:
So this change basically killed the throughput of the Kafka Producer since we have to individually wait for the RecordMetadata of each record being produced. This is now useless if you need to produce hundreds of thousands if not millions of records.
This is especially bad when your Kafka cluster and DebeziumServer are located in different data centers.

In the system I'm currently working on where we go from a local datacenter hosted SqlServer to AWS MSK, we are doing about ~70 records a second. This number needs to be north of 300.
In some use cases, we need to read from a SqlServer DB with millions of records in it which will now take weeks to complete whereas with the previous version, it would take a day or two at most.

debezium-server-kafka/src/main/java/io/debezium/server/kafka/KafkaChangeConsumer.java
RichardFourie: This `recordMetadataFuture.get`  for each record kills the throughput of the Kafka Producer.

AbriTj: We see lots of ASYNC_NETWORK_IO waits in SQL Server with this in place (https://www.sqlshack.com/reducing-sql-server-async_network_io-wait-type/). This did not happen in the previous version and it indicates that the SQL Server client, debezium server in this case, is slow in processing query results.

jpechane:
@RichardFourie Thanks for the rpeort! Could you please create Jira about the issue? We definitely need to switch to `send` method with async callback and gather results that way.

jpechane:
@RichardFourie Hi, I prepare a PR with fix (https://github.com/debezium/debezium-server/pull/123). Would you be able to build it and test if it solves the issue for you?

jpechane:
@AbriTj FYI ^^^

PR #117, opened 2024-08-29
DBZ-8186 Include MariaDB connector is distribution

Naros:
https://issues.redhat.com/browse/DBZ-8186


jpechane:
@Naros Applied, thanks

PR #118, opened 2024-09-02
DBZ-8101 Initial implementation of signal REST API for DS

jcechace:
https://issues.redhat.com/browse/DBZ-8101

### DIFF
--- a/debezium-server-core/pom.xml
+++ b/debezium-server-core/pom.xml
@@ -57,6 +57,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
             <scope>test</scope></dependency>
@@ -168,10 +178,11 @@
                 <configuration>
                     <skipTests>${skipITs}</skipTests>
                     <enableAssertions>true</enableAssertions>
-                    <systemProperties>
+                    <systemPropertyVariables>
                         <test.type>IT</test.type>
                         <test.apicurio>false</test.apicurio>
-                    </systemProperties>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                    </systemPropertyVariables>
                     <runOrder>${runOrder}</runOrder>
                 </configuration>
             </plugin>
--- a/debezium-server-core/src/main/java/io/debezium/server/DebeziumServerConfig.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DebeziumServerConfig.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server;
+
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+
+@ConfigMapping(prefix = "debezium")
+public interface DebeziumServerConfig {
+    Api api();
+
+    interface Api {
+        @WithDefault("false")
+        boolean enabled();
+    }
+}
--- a/debezium-server-core/src/main/resources/application.properties
+++ b/debezium-server-core/src/main/resources/application.properties
@@ -1,0 +1,3 @@
+debezium.api.enabled=false
+# TODO: this should be eventually removed when entire config is mapped
+smallrye.config.mapping.validate-unknown=false
--- a/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java
@@ -296,4 +296,8 @@ public class DebeziumServer {
     public Properties getProps() {
         return props;
     }
+
+    public DebeziumEngine.Signaler getSignaler() {
+        return engine.getSignaler();
+    }
 }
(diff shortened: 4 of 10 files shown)


jcechace:
@jpechane  since this introduces an insecure REST API, should we add some way of disabling/preventing access now or in a separate PR?

debezium-server-core/pom.xml
```diff
@@ -56,6 +56,16 @@
             <artifactId>quarkus-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jcechace, line 63: This dependency is going to be problematic it seems. It provides means to nicely test log output, however it also causes the DebeziumServerTest#testProps to fail with

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <"inventory.customers">
to be equal to:
 <"public.table_name">
but was not.
Expected :public.table_name
Actual   :inventory.customers
```

jcechace: Ok, false alarm the cause might be in failsafe configuration

debezium-server-core/pom.xml
```diff
@@ -168,10 +178,11 @@
                 <configuration>
                     <skipTests>${skipITs}</skipTests>
                     <enableAssertions>true</enableAssertions>
-                    <systemProperties>
+                    <systemPropertyVariables>
                         <test.type>IT</test.type>
                         <test.apicurio>false</test.apicurio>
-                    </systemProperties>
+                        <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                    </systemPropertyVariables>
```
jcechace, line 185: This is the problematic part. 

Prior to this change, integration tests don't run in IntelliJ idea  as it seems that the following error in IT results in no further output (and it looks like the test was not executed)

```
log4j:WARN No appenders could be found for logger (org.jboss.logging).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
Sep 02, 2024 4:03:52 PM org.jboss.logmanager.JBossLoggerFinder getLogger
ERROR: The LogManager accessed before the "java.util.logging.manager" system property was set to "org.jboss.logmanager.LogManager". Results may be unexpected.
```

jcechace:
Fixed the test failure, so everything should be passing now when run through maven. In IntelliJ the situation is reversed now -- while integration tests work, unit tests are broken.

vjuranek (commented):
LGTM, I'd just add a toggle to turn the (unsecure) rest api off by default as you already suggested.

jcechace:
@vjuranek well, let me know if you know how to do that. Otherwise the toggle is going to cause the API to simply return 305 as if signaler were not available

vjuranek:
> @vjuranek well, let me know if you know how to do that. Otherwise the toggle is going to cause the API to simply return 305 as if signaler were not available

I guess you mean 503 and IMHO it's fine (read: I don't know how to completely turn it off:-). Basically we just need to make sure Debezium server won't accept any signals unless it's explicitly configured to do so, with assumption the admin knows what he does when turning it on.

jcechace:
@vjuranek done, though we should probably consider disabled as the default value.

debezium-server-core/src/main/java/io/debezium/server/DebeziumServerConfig.java
```diff
@@ -0,0 +1,19 @@
...
+
+@ConfigMapping(prefix = "debezium")
+public interface DebeziumServerConfig {
+    Api api();
+
+    interface Api {
+        @WithDefault("false")
+        boolean enabled();
+    }
+}
```
jcechace, line 19: Eventually this should grow into a proper configuration mapping for the whole DS. It's about time we start slowing clearing up the implementation.

debezium-server-core/src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
+debezium.api.enabled=false
```
jpechane, line 1: Please see https://github.com/debezium/debezium-server/blob/main/debezium-server-dist/src/main/resources/distro/conf/application.properties.example

jpechane (commented):
@jcechace LGTM, thanks. I just don't think we should intorduce the new `application.properties` file if possible.

debezium-server-core/src/main/resources/application.properties
jcechace: Just for context. So, the problem is that by this file needs to be renamed and modified -- which means it can't really be  used to set defaults as user may easily omit them. 

However we do place this file on class path and thus run into a problem with classloader order -- although I believe that the `application.properties` inside the jar will be loaded first (and thus the one we add in startup script would be able to override the values inside it) it is not something we should be relying on. 

The main issue is that we shouldn't be adding the configuration file on cp in the first place. Quarkus (or rather smallrye config) will also automatically recognise `config/application.properties` and give it a higher precedence than cp config by default. However, to make things more complicated -- we don't just add the config file on cp, but rather the entire `conf` directory since while `application.properties` doesn't have to be on cp, things like `conf/cassandra/driver.conf` do.

jcechace: The discussed idea of not changing this at the moment and just relying on default in code will also not be feasible -- while it would generally work for this specific property, it pretty much means that we would not be able to do partial configuration mapping to a java class (which is cleaner than having bunch of `@ConfigProperty` randomly in the code) as that needs unknown property validation to be disabled (see the next line `smallrye.config.mapping.validate-unknown=false`). 

Thus as annoying as the change is, we do need to sort this out

jcechace: Opened #120 which needs to be merged prior to this

debezium-server-core/pom.xml
jcechace: @jpechane  any opinion regarding this?

jpechane: @jcechace If it works in CLI then I've no issue with it.

jpechane:
@jcechace Please rebase on the latest main and verify that the PR still works and if yes then feel free to merge yourself.

PR #119, opened 2024-09-09
[DBZ-7909] RabbitMQ Native Stream: Making sure RequestedFrameMax get sent to StreamClient

zikphil:
This fixes an issue I was having where I was setting the frame_max variable on the connection configuration but it was not propagating to the StreamClient. This patch fixes this issue.

https://issues.redhat.com/browse/DBZ-7909

### DIFF
--- a/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
+++ b/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
@@ -111,6 +111,7 @@ public class RabbitMqStreamNativeChangeConsumer extends BaseChangeConsumer imple
                     .username(factory.getUsername())
                     .password(factory.getPassword())
                     .virtualHost(factory.getVirtualHost())
+                    .requestedMaxFrameSize(factory.getRequestedFrameMax())
                     .build();
         }
         catch (StreamException | IllegalArgumentException e) {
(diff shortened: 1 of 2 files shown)


jpechane:
@zikphil Could you please change the commit message as it refers to DBZ-8201 instead of DBZ-7909

jpechane:
@zikphil LGTM, would it be possible to add the remaining config options of the `EnvironmentBuilder`?

zikphil:
> @zikphil LGTM, would it be possible to add the remaining config options of the `EnvironmentBuilder`?

@jpechane This is the part that is a bit unclear to be. **I do not code Java** but it appears to me that the ConnectionFactory is not used at all by the rabbitmq-stream-client so I am not sure why this is part of the debezium-server documentation and used in the class itself. Perhaps it is by convenience just to get validation on the host/port/virtualHost etc... but none of the wildcard will end up being used by the EnvironmentBuilder unless coded in.

I added a new commit that removes the usage of the ConnectionFactory and instead uses the variables documented here: [https://rabbitmq.github.io/rabbitmq-stream-java-client/stable/htmlsingle/#configuring-the-environment](https://rabbitmq.github.io/rabbitmq-stream-java-client/stable/htmlsingle/#configuring-the-environment)

I've also added some config for the StreamProducer defined here: [https://rabbitmq.github.io/rabbitmq-stream-java-client/stable/htmlsingle/#producer](https://rabbitmq.github.io/rabbitmq-stream-java-client/stable/htmlsingle/#producer)

It is a bigger change, perhaps we could squeeze it in for 3.0 Beta. If you accept this I can also put a PR for the documentation to be updated.

jpechane:
@zikphil Hi, I am sorry for the late reply. I think the PR is good. I'd just keep the old property names for host/port (with connection prefix), makr them as deprecate and emit a warning when used. They'd  be removed later. The docs update will reflect the new names and new config options. WDYT?

PR #123, opened 2024-09-20
DBZ-8251 Sent records without waiting

jpechane:
https://issues.redhat.com/browse/DBZ-8251


RichardFourie:
@jpechane Thanks for the fix. I might only be able to test it towards the end of next week but looking at the code it should have performance closer to the original `CountDownLatch` solution.

vjuranek (approved):

vjuranek:
Merge, thanks!

PR #126, opened 2024-09-25
DBZ-8236 Adjusted committer.markProcessed(record) to only execute after confirmed receipt of published messages by Pubsub

nathan-smit-1:
Hi there,

As discussed on jira ticket and here: https://debezium.zulipchat.com/#narrow/stream/350571-community-dbz-server/topic/Understanding.20record.2Ecommitter.20behaviour.20in.20Debezium.20server

I made the change in PubSubChangeConsumer as well as in PubSubLite as the same issue would occur there.  In this case, I wait until messageIds = ApiFutures.allAsList(deliveries).get(waitMessageDeliveryTimeout, TimeUnit.MILLISECONDS); completes before marking messages as processed.

From my testing, this resolves the issue.  Would it be possible also for this change to be added to the 2.7 branch?  Currently in Prod we would like to continue on the 2.7 and then just upgrade to 2.7.x.


jpechane:

Test failing test is passing locally, needs additional investigation.

PR #132, opened 2024-10-10
DBZ-8313 Added support for application.yaml in Debezium server by adding quarkus-config-yaml to POM

nathan-smit-1:
https://issues.redhat.com/browse/DBZ-8313


jpechane:
@nathan-smit-1 We have an properties base example in the distro. Do you think you can add a yaml based one as part of the PR too?

jpechane:

PR #134, opened 2024-10-28
DBZ-8367: fix error message: show the correct config-path & format

paumr:
When launching debezium-server via:
``` bash
java -jar ./debezium-server-dist/target/debezium-server-dist-3.0.0-SNAPSHOT-runner.jar io.debezium.server.Main
```

it errors out with the log-line:
```
Failed to load mandatory config value 'debezium.sink.type'. Please check you have a correct Debezium server config in /<redacted>/debezium/debezium-server/conf/application.properties or required properties are defined via system or environment variables.
```

This is broken in two ways:
- the printed path is wrong. quarkus loads application.properties from the `config` directory: https://quarkus.io/guides/config-reference
- the error message is logged via println. LOGGER should be used instead (otherwise depending on the logging-setup the message might be dropped - due to not being json)

This PR fixes both of these issues.

### DIFF
--- a/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java
@@ -278,9 +278,15 @@ public class DebeziumServer {
         }
         catch (NoSuchElementException e) {
             final String configFile = Paths.get(System.getProperty("user.dir"), "config", "application.properties").toString();
-            // CHECKSTYLE IGNORE check FOR NEXT 2 LINES
-            System.err.println(String.format("Failed to load mandatory config value '%s'. Please check you have a correct Debezium server config in %s or required "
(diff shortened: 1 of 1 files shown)


Naros:
Hi @paumr, can you please raise a Jira issue to track this change and update the PR commit messages to be prepended with the Jira issue key?

jpechane:
Hi, IIRC the `System.err` was used intentionally to make sure the error is printed even in case of logging misconfiguration.

paumr:
> Hi, IIRC the `System.err` was used intentionally to make sure the error is printed even in case of logging misconfiguration.

Hi, thanks for the information!
Maybe it would make sense to print the error both ways? That way we the message will never be dropped.

jpechane:
@paumr Makes sense

paumr:
As requested I:
- created a jira issue
- prepended all commits with it's ID (and the PR)
- modified the error message to print to stderr and LOGGER

jpechane:
@paumr Applied, thanks

PR #136, opened 2024-11-01
DBZ-8307 Fix loss data issue for RabbitMQ

Qwedir:
Done based on the principle of: https://github.com/debezium/debezium-server/pull/96

Link issue:
https://issues.redhat.com/browse/DBZ-8307

### DIFF
--- a/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
+++ b/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
@@ -10,6 +10,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
@@ -38,7 +41,6 @@ import io.debezium.server.BaseChangeConsumer;
  * Implementation of the consumer that delivers the messages into RabbitMQ Stream destination.
  *
  * @author Olivier Boudet
- *
  */
 @Named("rabbitmqstream")
 @Dependent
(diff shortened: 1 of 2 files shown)


Qwedir:
@jpechane I understand and see the issue. 
Here’s what I suggest as a solution. The code compiles, but I haven’t tested this solution.

debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
```diff
@@ -230,17 +235,41 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
         }
 
-        LOGGER.trace("Sent messages");
-        committer.markBatchFinished();
+        if (!hasError.get()) {
+            LOGGER.trace("All messages sent successfully");
+            committer.markBatchFinished();
+        }
+        else {
+            LOGGER.error("Batch processing was incomplete due to record processing errors.");
```
jpechane, line 271: I believe an exception should be thrown here.

debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
```diff
@@ -230,17 +235,41 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, RecordCommitt
...
+                            }
                         });
             }
             catch (StreamException e) {
                 throw new DebeziumException(e);
             }
+        }
 
-            committer.markProcessed(record);
+        if (!latch.await(producerConfirmTimeout, TimeUnit.SECONDS)) {
```
jpechane, line 261: IMHO we need a new config option for this wait set to a larger default value as timeout for sending a single message could be different to sending a batch.

jpechane (commented):
@Qwedir Nice work, I left two addiitional comments but I beleive the implementation as scuh is correct.

debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
Qwedir: done

debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamNativeChangeConsumer.java
Qwedir: done

PR #139, opened 2024-11-27
DBZ-8460 Add debezium-storage-jdbc to the distribution

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8460


jpechane:
@mfvitale Applied, thanks

PR #140, opened 2024-12-12
DBZ-3226: Add step to verify the Debezium-server distribution

Sgitario:
Added a new job to validate the distribution works for each of the supported sinks. For now, I only added the "redis" sink. 

Fixes https://issues.redhat.com/browse/DBZ-3226

### DIFF
--- a/.github/workflows/cross-maven.yml
+++ b/.github/workflows/cross-maven.yml
@@ -70,3 +70,63 @@ jobs:
         run: ./server/mvnw clean install -f core/pom.xml -DskipTests -DskipITs -Dformat.formatter.goal=validate -Dformat.imports.goal=check -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
       - name: Maven build Debezium Server
         run: ./server/mvnw clean install -fae -f server/pom.xml -Passembly -Dformat.formatter.goal=validate -Dformat.imports.goal=check -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 -DskipNonCore
+      - name: Archive Debezium distribution
+        uses: actions/upload-artifact@v4
+        with:
+          name: debezium-dist-artifact
+          path: ./server/debezium-server-dist/target/debezium-server-dist*.tar.gz
+
+  validate-distribution:
+    name: Verify debezium-server distribution
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sink: [ redis ]
+    needs: build
+    steps:
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 21
+      - name: Checkout Debezium Server repository
+        uses: actions/checkout@v4
+        with:
+          path: server
+      - name: Download Debezium distribution
+        uses: actions/download-artifact@v4
+        with:
+          name: debezium-dist-artifact
+      - name: Unzip Debezium distribution
+        run: tar xf debezium-server-dist-*.tar.gz
+      - name: Validate sink '${{matrix.sink}}'
+        run: |
+          # Creating the distribution archives
+          DISTRIBUTION_FOLDER=./debezium-server
+          SINK=${{matrix.sink}}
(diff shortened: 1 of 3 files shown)


.github/workflows/cross-maven.yml
```diff
@@ -70,3 +70,45 @@ jobs:
         run: ./server/mvnw clean install -f core/pom.xml -DskipTests -DskipITs -Dformat.formatter.goal=validate -Dformat.imports.goal=check -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
       - name: Maven build Debezium Server
         run: ./server/mvnw clean install -fae -f server/pom.xml -Passembly -Dformat.formatter.goal=validate -Dformat.imports.goal=check -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120 -DskipNonCore
+
+  validate-distribution:
```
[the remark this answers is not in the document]
jpechane: @Sgitario One is executed on pull requests. That job migh need to build core not from the `main` branch but from a different branch (same name as this one). This is used for cases when PR depends on changes in core repo.
`maven.yml` is used exclusively for main/branch builds and always build from `main` of the core repo

.github/workflows/cross-maven.yml
```diff
@@ -70,3 +70,53 @@ jobs:
...
```
jpechane, line 74: I wonder if a different job is needed as it executes yet another build. Can't the test be part of `build step?

Sgitario: I wanted to have a different job to use the strategy.matrix, so we can execute the same job to cover other different sinks / configurations with little effort.

Naros: I agree with Mario, if we can reuse the build across jobs since there is this _build_ dependency defined, I think that's reasonable to reduce the overall execution time on CI.

Sgitario: @mfvitale @Naros PR updated to reuse the distribution from the Build job. 
I confirmed that the validation job now takes a few seconds when before it used to take some minutes.

.github/workflows/cross-maven.yml
```diff
@@ -70,3 +70,63 @@ jobs:
...
+      matrix:
+        sink: [ redis ]
+    needs: build
+    steps:
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: 21
+      - name: Checkout Debezium Server repository
```
mfvitale, line 93: I think this is no more required right?

mfvitale (approved):

debezium-server-dist/src/test/resources/redis/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+# Docker file used by GitHub actions to verify the debezium-server-dist works.
+services:
+  postgres:
+    image: quay.io/debezium/example-postgres:2.1
```
jpechane, line 4:
```suggestion
    image: quay.io/debezium/example-postgres:3.0
```

jpechane (approved):
@Sgitario I think we've converged on good solution!
I left a comment WRT postgres exmaple version.

One mor eproposal for an imporvement - maybe a follow-up Jira. Would it be possible to connect to JMX as part of test evaluation and check metrics to make sure that expected number of messages was captured - effectivelly asserting https://debezium.io/documentation/reference/3.0/connectors/postgresql.html#connectors-snaps-metric-totalnumberofeventsseen_postgresql

debezium-server-dist/src/test/resources/redis/docker-compose.yml
Sgitario: Applied

Sgitario:
> One mor eproposal for an imporvement - maybe a follow-up Jira. Would it be possible to connect to JMX as part of test evaluation and check metrics to make sure that expected number of messages was captured - effectivelly asserting https://debezium.io/documentation/reference/3.0/connectors/postgresql.html#connectors-snaps-metric-totalnumberofeventsseen_postgresql

This sounds good +1. I've just created https://issues.redhat.com/browse/DBZ-8518 to add this. Also, as a follow-up task and when this functionality is more consolidated, we should also add the rest of the sinks to it. 

@jpechane I've just squashed all the commits.

jpechane:
@Sgitario Applied, thanks!

PR #150, opened 2025-01-16
V2.7.3.final nats publish async

ming-li-deel:
We found that sync publishing to nats has a very low throughput. We can only achieve 1k qps in production.
Proposing to publish to nats asynchronously to greatly improve the throughput.


ming-li-deel:
created by mistake, please ignore.

PR #155, opened 2025-02-03
Fix headers provided by HeaderConverter interface to http connector. …

oBJIADo:
I have tried to use debezium server with http-sink-connecter and i have faced 2 issues:

1. When headers are converted using the `HeaderConverter` interface (e.g., `SimpleHeaderConverter`), an error will always occur. This is because the converter, as defined by the interface, consistently returns a `byte[]`, whereas the `HttpChangeConsumer` only works with `String` headers.
example debezium props:
```
...
debezium.sink.type=http
...
debezium.format.header=clientprovided
debezium.format.header.class=org.apache.kafka.connect.storage.SimpleHeaderConverter
...
```

2. When the endpoint of the server is not reachable then debezium throws npe, because the message is null.

### DIFF
--- a/debezium-server-pravega/pom.xml
+++ b/debezium-server-pravega/pom.xml
@@ -10,7 +10,7 @@
   <name>Debezium Server Pravega Sink Adapter</name>
   <properties>
     <!-- IMPORTANT: This must be aligned with the netty shipped with Quarkus, specified in debezium-build-parent -->
-    <netty.version>4.1.108.Final</netty.version>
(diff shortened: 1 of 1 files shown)


jpechane:
@oBJIADo Hi, thanks for the PR! Could you please create a Jira issue describing the bug and then modify the commit message to `DBZ-xxxx Fix headers provided by HeaderConverter interface to...`?
We want to make sure it is correctly tracked. Also that way it will be backported to 3.0 branch too.

PR #156, opened 2025-02-06
DBZ-8634 Milvus sink

jpechane:


vjuranek (commented):
I only skimmed through the PR, so maybe I'm missing something, but why is all the Pulsar and related stuff (bookkeeper, zookeeper) needed? Or all this stuff will be removed (as a remnant from copying Pulsar module)?

jpechane:
@vjuranek The PR was cleaned and the implementation is at MVP level. Please take a look.
Later on, when your SMT is finished we can add another test that would take a table without vector column and add the vector via the vectorization SMT.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
```diff
@@ -0,0 +1,207 @@
...
+                              RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        final var data = getValue(record, toSourceRecord(record));
+
+        final var request = UpsertReq.builder()
+                .collectionName(collectionName)
+                .data(List.of(data))
+                .build();
+        milvusClient.upsert(request);
+        committer.markProcessed(record);
```
vjuranek, line 169: shouldn't we check `upsert()` result before marking record as processed (same for delete)?

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private void upsertRecord(String collectionName, ChangeEvent<Object, Object> record,
+                              RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        final var data = getValue(record, toSourceRecord(record));
+
+        final var request = UpsertReq.builder()
+                .collectionName(collectionName)
+                .data(List.of(data))
```
vjuranek, line 166: In the future we can possibly add caching of the records from given batch and do the bulk update.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusSchema.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Manages collections in Milvus database. Maps Kafka Connect schema into Milvus schema and validates
+ * contraints.
+ *
+ * @author Jiri Pechanec
+ */
+public class MilvusSchema {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MilvusChangeConsumer.class);
```
vjuranek, line 34:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(MilvusChangeConsumer.class);
    private static final Logger LOGGER = LoggerFactory.getLogger(MilvusSchema.class);
```

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
```diff
@@ -0,0 +1,207 @@
...
+                continue;
+            }
+
+            if (record.key() == null) {
+                throw new DebeziumException("Milvus does not support collections without primary key");
+            }
+            schema.validateKey(collectionName, sourceRecord.keySchema());
+
+            if (sourceRecord.value() == null) {
+                deleteRecord(collectionName, record, committer);
```
vjuranek, line 124: Do we actually need to handle tombstones? The record should be deleted upon receiving delete event.

vjuranek (commented):
Few minor comments, but over LGTM, nice work!

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
jpechane: I agree for starter I just intend to keep things simple and stupid.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
jpechane: Good point, when I was doing some testing during RnD phase I was getting exceptions for which this is safe approach but definitely better to check the response too.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
jpechane: This is intentional in case `ExtractNewRecordState` SMT is applied. As soon as we agree on handingld of deletes/tombstones in JDBC sink for different siatiations I'll apply the same logic.
Currently I just took the most conservative approach even at the cost of double calls on delete.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
jpechane: Well, I checked the API and it seems there is no status retunred, only the number of records modified. In this case I think we can ignore the results as zero result can be valid in some situations too.

jpechane (commented):
@vjuranek Please check the answers

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
vjuranek: I thought it [returns status](https://github.com/milvus-io/milvus-sdk-java/blob/master/sdk-core/src/main/java/io/milvus/client/MilvusClient.java#L414), but I didn't check it, so yes, in such case this is completely fine.

debezium-server-milvus/src/main/java/io/debezium/server/milvus/MilvusChangeConsumer.java
vjuranek: I see, I forgot about `ExtractNewRecordState`

vjuranek (approved):
LGTM (once the compilation failure is fixed)

PR #157, opened 2025-02-20
DBZ-8715 Added message compression and concurrency control

nathan-smit-1:
Hi there,

Made adjustments to include concurrency and compression to pub/sub:

https://cloud.google.com/pubsub/docs/concurrency-control-messages
https://cloud.google.com/pubsub/docs/compress-messages

Set the default concurrency threads to 1 as I imagine anyone setting concurrency to true has some specific goal in mind and will almost certainly adjust.  The compression threshold I set is the default value as per the docs.

### DIFF
--- a/debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
+++ b/debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java
@@ -74,6 +74,12 @@ public class PubSubChangeConsumer extends BaseChangeConsumer implements Debezium
     private static final String PROP_PREFIX = "debezium.sink.pubsub.";
     private static final String PROP_PROJECT_ID = PROP_PREFIX + "project.id";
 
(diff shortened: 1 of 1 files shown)


nathan-smit-1:
I'm not sure what this build failure is but I'm seeing it on some of the other pull requests as well

jpechane:
@nathan-smit-1 Thanks for the PR. I wonder if in both cases do we need the enabled config option. Maybe if we introduce `0` as a special value for number of threads and `-1` for compression threshold then we could use them as enabled/disabled flag. This would reduce the total number of config options. WDYT?

nathan-smit-1:
> @nathan-smit-1 Thanks for the PR. I wonder if in both cases do we need the enabled config option. Maybe if we introduce `0` as a special value for number of threads and `-1` for compression threshold then we could use them as enabled/disabled flag. This would reduce the total number of config options. WDYT?

@jpechane I think that makes sense.  Are you okay with the implementation I just pushed with just the two configs?  I also updated documentation on this pull:  https://github.com/debezium/debezium/pull/6176

jpechane:

PR #159, opened 2025-02-24
DBZ-8672 Don't fail server shutdown

vjuranek:
Don't fail server shutdown if engine shutdown is already in progress Throwing exception from engine `close()` method would result into skipping shutting down the executor, which is not nice.

https://issues.redhat.com/browse/DBZ-8672


jpechane:
@vjuranek Applied, thanks!

PR #161, opened 2025-03-07
DBZ-8752 New property rabbitmq.routingKey.source for static, topic an…

victorcastanogutierrez:
…d key states

Related issue: https://issues.redhat.com/browse/DBZ-8752

This PR adds support for a new property on the RabbitMQ sink `debezium.sink.rabbitmq.routingKey.source`. The motivation is to allow not only producing to desired exchange based on the outbox table, but also being able to dynamically select the routing key from the same table.

For that, this new property has the following possible values:

- `static` (default): which will obtain the routing key from `debezium.sink.rabbitmq.routingKey`
- `topic`: uses the value of the destination exchange as the routing key
- `key`: uses the record key as the routing key

`debezium.sink.rabbitmq.routingKeyFromTopicName` is now deprecated as `debezium.sink.rabbitmq.routingKey.source` with `topic` value does the same thing. If set to true, a warn will be emitted and the value of `debezium.sink.rabbitmq.routingKey.source` will become `topic` in order to provide backwards compatibility

### DIFF
--- a/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamChangeConsumer.java
+++ b/debezium-server-rabbitmq/src/main/java/io/debezium/server/rabbitmq/RabbitMqStreamChangeConsumer.java
@@ -31,11 +31,13 @@ import com.rabbitmq.client.ConnectionFactory;
 import com.rabbitmq.client.ConnectionFactoryConfigurator;
 
 import io.debezium.DebeziumException;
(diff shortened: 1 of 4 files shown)


victorcastanogutierrez:
Sure thing, will do it as soon as I can

victorcastanogutierrez:
Thank you!

victorcastanogutierrez:
Docs PR https://github.com/debezium/debezium/pull/6219

PR #163, opened 2025-03-13
DBZ-8637 InstructLab QnA sink

Naros:
https://issues.redhat.com/browse/DBZ-8637


debezium-server-instructlab/src/main/java/io/debezium/server/instructlab/InstructLabSinkConsumer.java
```diff
@@ -0,0 +1,113 @@
...
+        Headers kafkaHeaders = event.sourceRecord().headers();
+        for (Header header : kafkaHeaders) {
+            final String key = header.key();
+            if (key.startsWith(AbstractAddQnaHeader.QNA_PREFIX)) {
+                String[] keyParts = key.split("\\.", 3);
+                if (keyParts.length == 3) {
+                    final String prefix = keyParts[1];
+                    final String attribute = keyParts[2];
+                    headers.computeIfAbsent(prefix, k -> new HashMap<>()).put(attribute, String.valueOf(header.value()));
+                }
```
vjuranek, line 107: Maybe add `else` branch with printing debug message that key has wrong format and record will be skipped?

debezium-server-instructlab/src/main/java/io/debezium/server/instructlab/InstructLabSinkConsumer.java
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                });
+            }
+            committer.markProcessed(record);
+        }
+
+        // Flush files
+        for (QnaFile file : batchFiles.values()) {
+            try {
+                file.flush();
```
vjuranek, line 84: As the disk write operation may be slow, in the future we may add an option to flush files in async manner (with the risk that data may be lost).

vjuranek (approved):
Just one minor thing, otherwise LGTM. Great work @Naros !

debezium-server-instructlab/src/main/java/io/debezium/server/instructlab/InstructLabSinkConsumer.java
Naros: Added a new log entry as requested, LMK if that fits the bill.

vjuranek: yes, it does

jpechane (commented):
@Naros Nice work, thanks!
I left one comment related to taxonomy header name. Still I beleive we focused on headers a bit too much. What if question, answer and taxnomy would come directly from the database? Can't we modify the sink in the sense we do in other SMTs/convertes/... where the source of the data could be directly the field or header based on for example `header:`/`value:` prefix?

Naros:
> What if question, answer and taxnomy would come directly from the database? Can't we modify the sink in the sense we do in other SMTs/convertes/... where the source of the data could be directly the field or header based on for example `header:`/`value:` prefix?

Sure. I would assume we'd make the sink more configuration-driven, i.e.:
```
debezium.sink.instructlab.taxonomy.path=<path to the top-level taxonomy location>
debezium.sink.instructlab.mappings=a,b,c
debezium.sink.instructlab.mappings.a.question=header:h1
debezium.sink.instructlab.mappings.a.answer=header:h2
debezium.sink.instructlab.mappings.a.taxonomy=header:h3
debezium.sink.instructlab.mappings.b.question=value:B_QUESTION
debezium.sink.instructlab.mappings.b.answer=value:B_ANSWER
debezium.sink.instructlab.mappings.b.context=value:B_CONTEXT
debezium.sink.instructlab.mappings.b.taxonomy=value:B_TAXONOMY
debezium.sink.instructlab.mappings.c.question=value:C_QUESTION
debezium.sink.instructlab.mappings.c.answer=value:C_ANSWER
debezium.sink.instructlab.mappings.c.taxonomy=/domain/part-a/part-b
```

In this configuration example:

* SMTs can add whatever headers with values. The mappings, if they are prefixed with `header:` will resolve the header key from the event and use the value from the header for that mapping.
* Databases can populate fields with values. The mappings, if they are prefixed with `value:` will resolve to that field. We can either allow mapping nested names like `after.field` or just `field` and then resolve whether to prepend `after` if its a Debezium event versus a flattened event.
* Values can be sourced statically from the configuration by not pretending `header:` or `value:` to the mappings. 

In addition, the `<mapping>.taxonomy` will expect a slash-based set of segments that represent the parts that make up the taxonomy path. The `instructlab.taxonomy.path` will be appended by the slash-segments followed by `qna.yml`.

This reduces the dependency on headers as you want and it also allows the SMTs to be decoupled from the sink's internal logic and forces everything to be more configuration-driven.  This also allows for the same idea of mix and matching various transforms, database fields, and static values.

Is this more inline with what you had in mind @jpechane ?

jpechane:
@Naros Perfect, it is even more than I epxected!

PR #178, opened 2025-04-29
DBZ-8985 Don't use wildcards in stream subjects

vjuranek:
If the subjects are not specified, `*.*.*` is used.

https://issues.redhat.com/browse/DBZ-8985


jpechane:
@vjuranek Applied, thanks

PR #193, opened 2025-07-01
Implement dynamic partition key support for Azure Event Hub sink

lwu-st:
When no explicit partitionId or partitionKey was configured, the system was using `record.partition()` which could be null, and then falling back to `BATCH_INDEX_FOR_NO_PARTITION_ID` (-1)
This caused all messages to have null partition IDs.

This PR modifies the partition selection logic to use `record.key()` as the partition key when no explicit partitioning is configured. This matches the behavior of other sinks like Kinesis, which already use `record.key()`

### DIFF
--- a/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
+++ b/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
@@ -29,7 +29,7 @@ public class BatchManager {
     // Prepare CreateBatchOptions for N partitions
     private final HashMap<Integer, CreateBatchOptions> batchOptions = new HashMap<>();
     private final HashMap<Integer, EventDataBatchProxy> batches = new HashMap<>();
-    
+
     private final HashMap<String, EventDataBatchProxy> dynamicPartitionKeyBatches = new HashMap<>();
 
     public BatchManager(EventHubProducerClient producer, String configurePartitionId,
@@ -95,7 +95,7 @@ public class BatchManager {
                 emitBatchToEventHub(batch);
             }
         });
-        
+
         dynamicPartitionKeyBatches.forEach((partitionKey, batch) -> {
             if (batch.getCount() > 0) {
                 LOGGER.trace("Dispatching {} events for partition key '{}'.", batch.getCount(), partitionKey);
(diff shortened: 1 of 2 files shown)


debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
```diff
@@ -121,6 +130,49 @@ public void sendEventToPartitionId(EventData eventData, Integer recordIndex, Int
         }
     }
 
+    /**
+     * Sends an event using a dynamic partition key derived from the record key.
+     * This allows proper partitioning when no explicit partition ID or key is configured.
+     */
+    public void sendEventWithDynamicPartitionKey(EventData eventData, String partitionKey) {
+        String effectivePartitionKey = (partitionKey != null && !partitionKey.isEmpty()) ? partitionKey : "default";
```
vjuranek:
```suggestion
        String effectivePartitionKey = Strings.isNullOrBlank((partitionKey) ? partitionKey : "default";
```

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
```diff
@@ -164,21 +165,31 @@ else if (!configuredPartitionKey.isEmpty()) {
                     targetPartitionId = BatchManager.BATCH_INDEX_FOR_PARTITION_KEY;
                 }
                 else {
-                    targetPartitionId = record.partition();
-
-                    if (targetPartitionId == null) {
-                        targetPartitionId = BatchManager.BATCH_INDEX_FOR_NO_PARTITION_ID;
+                    if (record.key() != null) {
```
vjuranek, line 168: This would break backward compatibility. I wonder if this is fine or we should introduce new option (which needs to be turned on by the user) to enable it. @jpechane WDYT?

lwu-st: Do you think that there's a case where this would cause problems? 

I assume up until now, no one who's used Az Event Hub has ever had a partitioned topic, because they would have noticed that messages only go into 1 partition. 
Similarly, I don't think anyone has ever had a log compacted topic, because they would have seen the connector crash on startup, or be forced to set a static `partitionKey`, causing compaction to shrink their topic to (1) message.

vjuranek: Well, I wasn't sure if this is an issue or not, but this is a good point, so IMHO we are good to go as it is.

vjuranek (approved):

vjuranek:
LGTM, but let's wait for CI before merging.

vjuranek:
I fixed compilation error and also added issue prefix to remaining commit messages.

vjuranek:
Merged, thanks @lwu-st !

PR #195, opened 2025-07-03
DBZ-9195: Fix inverted partitionKey coalesce for Event Hub sink

lwu-st:


Naros (approved):
LGTM, once CI passes we can merge 👍

PR #202, opened 2025-08-07
DBZ-9304 Clear dynamic batches between batch calls

jpechane:
https://issues.redhat.com/browse/DBZ-9304

### DIFF
--- a/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
+++ b/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
@@ -6,6 +6,7 @@
 package io.debezium.server.eventhubs;
 
 import java.util.HashMap;
+import java.util.Map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -19,6 +20,7 @@ import io.debezium.util.Strings;
 
 public class BatchManager {
     private static final Logger LOGGER = LoggerFactory.getLogger(BatchManager.class);
+
     private final EventHubProducerClient producer;
     private final String configuredPartitionId;
     private final String configuredPartitionKey;
@@ -28,10 +30,10 @@ public class BatchManager {
     static final Integer BATCH_INDEX_FOR_PARTITION_KEY = 0;
 
     // Prepare CreateBatchOptions for N partitions
-    private final HashMap<Integer, CreateBatchOptions> batchOptions = new HashMap<>();
-    private final HashMap<Integer, EventDataBatchProxy> batches = new HashMap<>();
+    private final Map<Integer, CreateBatchOptions> batchOptions = new HashMap<>();
+    private final Map<Integer, EventDataBatchProxy> staticBatches = new HashMap<>();
 
-    private final HashMap<String, EventDataBatchProxy> dynamicPartitionKeyBatches = new HashMap<>();
+    private final Map<String, EventDataBatchProxy> dynamicPartitionKeyBatches = new HashMap<>();
 
     public BatchManager(EventHubProducerClient producer, String configurePartitionId,
                         String configuredPartitionKey, Integer maxBatchSize) {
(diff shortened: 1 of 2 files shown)


Naros (approved):
This looks reasonable to me @jpechane :+1:

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
```diff
@@ -101,8 +103,20 @@ public void closeAndEmitBatches() {
             if (batch.getCount() > 0) {
                 LOGGER.trace("Dispatching {} events for partition key '{}'.", batch.getCount(), partitionKey);
                 emitBatchToEventHub(batch);
+
+                // Clear the batch after emitting to avoid re-sending the same events
+                // This is important for dynamic partition keys as they are created on-the-fly.
+                // The code is commented out as it can potentially lead to increased memory usage
+                // if the partition key is not re-used
+                // batch.clear();
```
vjuranek, line 117: Is this really needed? Batches are re-created at the beginning of each Debezium batch iteration.

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
```diff
@@ -101,8 +103,20 @@ public void closeAndEmitBatches() {
...
             }
         });
+
+        // Clear all dynamic batches to avoid emitting them again
+        // If it proves that batch re-initialization is too expensive for re-used keys
+        // we might need to think of using BoundedConcurrentHashMap but it would be necessary to ensure
+        // that its size is larger then the maximum number of changes coming in batch.
+        dynamicPartitionKeyBatches.clear();
```
vjuranek: I'd move at the end of `BatchManager#initializeBatch` method where non-dynamic batches are re-created as this has similar functionality - to start with clear batches (and let needed dynamic batches to be re-created on the fly as it is now).

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
jpechane: @vjuranek Only statically configured batches, not the dynamic ones. This means that the dynamic ones are kept from previous run.

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
jpechane: @vjuranek Good point, moved

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/BatchManager.java
vjuranek: Yes, but with change in this PR (`dynamicPartitionKeyBatches.clear();`), dynamic batches are removed at the beginning of each Debezium batch iteration, so doing the cleanup of internal batches at the end of the iteration seems to me redundant.

jpechane: Please see that this is commented code - it is more like alternative approach if the current one will not work under certain conditions.

vjuranek: Yes, I noticed that, but find it confusing - not now, but say in a year I can imagine myself scratching my head why the code which does the cleanup and prevents duplicates is commented out. So I'd at least extend the comment with explanation that the actual cleanup is done in `initializeBatch()` and this is another way how to do it if needed or something similar.

jpechane: @vjuranek God point, comment extended

vjuranek (approved):
Thanks for the patience @jpechane , now LGTM :-)

PR #206, opened 2025-08-12
[ci] Use the same sanity check as main Debezium repo

vjuranek:


Naros (approved):
Thanks @vjuranek, applied!

PR #212, opened 2025-09-15
DBZ-9468 Don't exclude maven-artifact artifact

vjuranek:
There is only one Apache Maven dependency, `maven-artifact` artifact, which is now required by Kafka 4.1.0.

https://issues.redhat.com/browse/DBZ-9468


jpechane:
Do we need `plexus`?

jpechane:
@vjuranek Applied, thanks

PR #213, opened 2025-09-18
DBZ-9477: support partition-scoped SQL Server connector metrics

kinolaev:
Partial solution for https://issues.redhat.com/browse/DBZ-9477.
The problem is SQL Server connector metrics are split between several mbeans ([task- and partition-scoped](https://github.com/debezium/debezium/blob/v3.3.0.Beta1/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/metrics/SqlServerStreamingTaskMetrics.java)).
This PR is not a general solution for the problem but it allows to use DebeziumMetrics when SQL Server connector is configured with 1 task and 1 database (partition) and should not affect other connectors.

### DIFF
--- a/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
@@ -37,7 +37,7 @@ public class DebeziumMetrics {
     private ObjectName streamingMetricsObjectName;
     private ObjectName streamingPartitionMetricsObjectName;
 
-    private static ObjectName getDebeziumMbean(String context, boolean partition) {
+    private static ObjectName getDebeziumMbean(String context, boolean partitioned) {
         ObjectName debeziumMbean = null;
 
         for (ObjectName mbean : mbeanServer.queryNames(null, null)) {
@@ -46,7 +46,7 @@ public class DebeziumMetrics {
                     && mbean.getCanonicalName().contains("type=connector-metrics")
                     && mbean.getCanonicalName().contains("context=" + context)
                     && (mbean.getCanonicalName().contains("debezium.sql_server:")
-                            ? mbean.getCanonicalName().contains("database=") == partition
+                            ? mbean.getCanonicalName().contains("database=") == partitioned
                             : true)) {
                 LOGGER.debug("Using {} MBean to get {} metrics", mbean, context);
                 debeziumMbean = mbean;
--- a/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
@@ -33,16 +33,21 @@ public class DebeziumMetrics {
     public static final MBeanServer mbeanServer = ManagementFactory.getPlatformMBeanServer();
 
     private ObjectName snapshotMetricsObjectName;
+    private ObjectName snapshotPartitionMetricsObjectName;
     private ObjectName streamingMetricsObjectName;
+    private ObjectName streamingPartitionMetricsObjectName;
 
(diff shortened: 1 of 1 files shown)


debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -58,21 +63,39 @@ private static ObjectName getDebeziumMbean(String context) {
...
+        }
+
+        return snapshotPartitionMetricsObjectName;
+    }
+
     public ObjectName getStreamingMetricsObjectName() {
 
         if (streamingMetricsObjectName == null) {
-            streamingMetricsObjectName = getDebeziumMbean("streaming");
+            streamingMetricsObjectName = getDebeziumMbean("streaming", false);
```
vjuranek, line 89: Isn't `QueueTotalCapacity` in `SqlServerStreamingTaskMetrics`, which is partitioned, so shouldn't here be `true`?

debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -33,16 +33,21 @@ public class DebeziumMetrics {
     public static final MBeanServer mbeanServer = ManagementFactory.getPlatformMBeanServer();
 
     private ObjectName snapshotMetricsObjectName;
+    private ObjectName snapshotPartitionMetricsObjectName;
     private ObjectName streamingMetricsObjectName;
+    private ObjectName streamingPartitionMetricsObjectName;
 
-    private static ObjectName getDebeziumMbean(String context) {
+    private static ObjectName getDebeziumMbean(String context, boolean partition) {
```
vjuranek: maybe use `partitioned` instead of `partition`?

debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
kinolaev: Yes, `QueueTotalCapacity` is in `SqlServerStreamingTaskMetrics` and not in `SqlServerStreamingPartitionMetrics`, so it is not partitioned

debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
kinolaev: Done in 6d236e1eca2fd59bef033555db20a49ca9877bd3

debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
vjuranek: ah, right, sorry

vjuranek (approved):

debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -33,16 +33,21 @@ public class DebeziumMetrics {
...
+    private static ObjectName getDebeziumMbean(String context, boolean partitioned) {
         ObjectName debeziumMbean = null;
 
         for (ObjectName mbean : mbeanServer.queryNames(null, null)) {
 
             if (mbean.getCanonicalName().contains("debezium.")
                     && mbean.getCanonicalName().contains("type=connector-metrics")
-                    && mbean.getCanonicalName().contains("context=" + context)) {
+                    && mbean.getCanonicalName().contains("context=" + context)
+                    && (mbean.getCanonicalName().contains("debezium.sql_server:")
```
jpechane: Could we extract this sub-condition to a separate method with suitable to better convey its meaning

vjuranek (approved):

PR #215, opened 2025-09-19
DBZ-9474 Adds support for publishAsync to nats jetstream sink

notation-v:
Adds support for publishAsync to nats jetstream sink

- Implements an async method to nats jetstream sink.

### DIFF
--- a/debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
+++ b/debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
@@ -11,6 +11,10 @@ import java.security.KeyStore;
 import java.security.SecureRandom;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
@@ -38,11 +42,14 @@ import io.debezium.server.CustomConsumerBuilder;
 import io.nats.client.Connection;
 import io.nats.client.JetStream;
 import io.nats.client.JetStreamManagement;
+import io.nats.client.Message;
 import io.nats.client.Nats;
 import io.nats.client.Options;
+import io.nats.client.api.PublishAck;
 import io.nats.client.api.StorageType;
 import io.nats.client.api.StreamConfiguration;
 import io.nats.client.impl.Headers;
+import io.nats.client.impl.NatsMessage;
 
 /**
  * Implementation of the consumer that delivers the messages into a NATS Jetstream stream.
@@ -70,6 +77,9 @@ public class NatsJetStreamChangeConsumer extends BaseChangeConsumer
     private static final String PROP_AUTH_TLS_KEYSTORE_PASSWORD = PROP_PREFIX + "auth.tls.keystore.password";
     private static final String PROP_AUTH_TLS_PASSWORD = PROP_PREFIX + "auth.tls.password";
 
+    private static final String PROP_ASYNC_ENABLED = PROP_PREFIX + "async.enabled";
+    private static final String PROP_ASYNC_TIMEOUT_MS = PROP_PREFIX + "async.timeout-ms";
+
     private Connection nc;
     private JetStream js;
 
@@ -96,6 +106,12 @@ public class NatsJetStreamChangeConsumer extends BaseChangeConsumer
 
     @ConfigProperty(name = PROP_AUTH_TLS_PASSWORD)
     Optional<String> tlsPassword;
+
+    @ConfigProperty(name = PROP_ASYNC_ENABLED, defaultValue = "true")
+    boolean asyncEnabled;
+
+    @ConfigProperty(name = PROP_ASYNC_TIMEOUT_MS, defaultValue = "5000")
+    long asyncTimeoutMs;
 
     @Inject
     @CustomConsumerBuilder
@@ -158,6 +174,13 @@ public class NatsJetStreamChangeConsumer extends BaseChangeConsumer
         catch (Exception e) {
             throw new DebeziumException(e);
         }
+
+        if (asyncEnabled) {
+            LOGGER.info("Async publishing enabled with timeout: {}ms", asyncTimeoutMs);
+        }
+        else {
+            LOGGER.info("Synchronous publishing mode enabled");
+        }
     }
 
     @PreDestroy
@@ -176,6 +199,16 @@ public class NatsJetStreamChangeConsumer extends BaseChangeConsumer
     @Override
     public void handleBatch(List<ChangeEvent<Object, Object>> records,
                             RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        if (!asyncEnabled || records.isEmpty()) {
+            handleBatchSync(records, committer);
+            return;
+        }
+        handleBatchAsync(records, committer);
+    }
+
+    private void handleBatchSync(List<ChangeEvent<Object, Object>> records,
+                                 RecordCommitter<ChangeEvent<Object, Object>> committer)
             throws InterruptedException {
 
         for (ChangeEvent<Object, Object> rec : records) {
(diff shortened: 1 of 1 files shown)


debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
```diff
@@ -70,6 +77,9 @@ public class NatsJetStreamChangeConsumer extends BaseChangeConsumer
     private static final String PROP_AUTH_TLS_KEYSTORE_PASSWORD = PROP_PREFIX + "auth.tls.keystore.password";
     private static final String PROP_AUTH_TLS_PASSWORD = PROP_PREFIX + "auth.tls.password";
 
+    private static final String PROP_ASYNC_ENABLED = PROP_PREFIX + "async.enabled";
+    private static final String PROP_ASYNC_TIMEOUT_MS = PROP_PREFIX + "async.timeout-ms";
```
jpechane:
```suggestion
    private static final String PROP_ASYNC_TIMEOUT_MS = PROP_PREFIX + "async.timeout.ms";
```

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
```diff
@@ -211,6 +244,69 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records,
...
+                            .subject(subject)
+                            .headers(natsHeaders)
+                            .data(recordBytes)
+                            .build();
+
+                    LOGGER.trace("Sending async message to subject: {} (data length: {} bytes)", subject, recordBytes.length);
+
+                    CompletableFuture<PublishAck> future = js.publishAsync(msg).orTimeout(asyncTimeoutMs, TimeUnit.MILLISECONDS);
+
+                    future.thenAccept(ack -> {
```
jpechane: Maybe this should be guarded with `if (LOGGER.isTraceEnabled)`?

jpechane:
```suggestion
                    future = future.thenAccept(ack -> {
```

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
```diff
@@ -211,6 +244,69 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records,
...
+                    .get(asyncTimeoutMs, TimeUnit.MILLISECONDS);
+
+            for (ChangeEvent<Object, Object> rec : records) {
+                committer.markProcessed(rec);
+            }
+            committer.markBatchFinished();
+            LOGGER.debug("Successfully published batch of {} messages", records.size());
+        }
+        catch (TimeoutException e) {
+            LOGGER.error("Timeout waiting for publish acknowledgments after {}ms", asyncTimeoutMs);
```
jpechane:
```suggestion
            LOGGER.error("Timeout waiting for publish acknowledgments after {} ms", asyncTimeoutMs);
```

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
```diff
@@ -211,6 +244,69 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records,
...
+                .toList();
+        if (futures.isEmpty()) {
+            for (ChangeEvent<Object, Object> rec : records) {
+                committer.markProcessed(rec);
+            }
+            committer.markBatchFinished();
+            return;
+        }
+        try {
+            CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
```
jpechane:
```suggestion
            CompletableFuture.allOf(futures.toArray())
```

jpechane (commented):
@notation-v Thanks for the PR! I closed the other one in favour of this. There are few minor comments left. Coudl you please take a look at them? I believe we are done when addressed.

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
notation-v: not much to compute around it, so we should be good without a guard.

notation-v: Hello @jpechane, any reasons why it should be this? `.thenAccept` returns a `CompletableFuture<void>`, can't be assigned to future of type `CompletableFuture<PublishAck>`

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
notation-v: Object[] won't resolve to CompletableFuture, so this casting is required.

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
jpechane: `thenAccept` creates and return new `CompletableFuture` so the new one should be used, shouldn't it?

jpechane: As mentnioned below, `thenAccept` creates new `CompletableFuture`. This is critical path of the code executed for every single message so I'd prefer to do it only when needed.

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
jpechane: In that case `new CompletableFuture[futures.size()]` could be used

notation-v: the JDK doc recommends `new CompletableFuture[0]` as more idiomatic implementation, my IDE also warns me of it. but if you really want to define the size like this, I will push the change.

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
notation-v: `thenAccept` returned future (`CompletableFuture<void>`) is for logging, I think it is an anti-pattern to track the logging future, which is a side-effect of `publishAsync`, instead of the future that is responsible for `publishAsync`. If the logging fails for some reasons, do we want to consider it a publish failure?

jpechane: Here I think we misundertood each other, I'd add complete `future.thenAccept` call into `if (LOGGER.isTraceEnabled())` so the new stage is not created at all.

jpechane: OK, makes sense

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
jpechane: In this case I refer to this part of JavaDoc
```
If the list fits in the specified array, it is returned therein. Otherwise, a new array is allocated with the runtime type of the specified array and the size of this list.
```
also I don't see it recommended as idiomatic. From my PoV if one knows the size of array in advance it should be passed directly.

notation-v: got it, I have made this change as per recommendation. javadoc for `toArray()` has an example to letting it auto assign.

debezium-server-nats-jetstream/src/main/java/io/debezium/server/nats/jetstream/NatsJetStreamChangeConsumer.java
notation-v: done

jpechane:
@notation-v Applied, thanks a lot! Could you please send a PR to the main repo with docs update too?

PR #216, opened 2025-09-24
DBZ-9059 Enable OpenLineage output dataset events on Debezium Server

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9059

image

### DIFF
--- a/debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server;
+
+import static io.debezium.openlineage.OpenLineageConfig.OPEN_LINEAGE_INTEGRATION_ENABLED;
+import static io.debezium.openlineage.dataset.DatasetMetadata.STREAM_DATASET_TYPE;
+import static io.debezium.openlineage.dataset.DatasetMetadata.DataStore.KAFKA;
+import static io.debezium.openlineage.dataset.DatasetMetadata.DatasetKind.OUTPUT;
+import static io.debezium.server.DebeziumServer.PROP_SINK_TYPE;
+import static io.debezium.server.DebeziumServer.PROP_SOURCE_PREFIX;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import org.apache.kafka.connect.header.ConnectHeaders;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.eclipse.microprofile.config.Config;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.DebeziumException;
+import io.debezium.connector.common.DebeziumTaskState;
+import io.debezium.embedded.EmbeddedEngineChangeEvent;
+import io.debezium.engine.ChangeEvent;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.openlineage.ConnectorContext;
+import io.debezium.openlineage.DebeziumOpenLineageEmitter;
+import io.debezium.openlineage.dataset.DatasetDataExtractor;
+import io.debezium.openlineage.dataset.DatasetMetadata;
+
+public class DefaultChangeConsumer extends BaseChangeConsumer implements DebeziumEngine.ChangeConsumer<ChangeEvent<Object, Object>> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultChangeConsumer.class);
+
+    private final DebeziumEngine.ChangeConsumer<ChangeEvent<Object, Object>> delegateConsumer;
+    private final Config config;
+    private final DatasetDataExtractor datasetDataExtractor;
+
+    public DefaultChangeConsumer(DebeziumEngine.ChangeConsumer<ChangeEvent<Object, Object>> delegateConsumer, Config config) {
+        this.delegateConsumer = delegateConsumer;
+        this.config = config;
+        datasetDataExtractor = new DatasetDataExtractor();
+    }
+
+    @Override
+    public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+
+        final String sink = config.getValue(PROP_SINK_TYPE, String.class);
+
+        try {
+            this.delegateConsumer.handleBatch(records, committer);
+        }
+        catch (Exception e) {
+            throw new DebeziumException("Error while executing batch", e);
+        }
+
+        Boolean isOpenLineageEnabled = config.getValue(PROP_SOURCE_PREFIX + OPEN_LINEAGE_INTEGRATION_ENABLED, Boolean.class);
+        if (isOpenLineageEnabled) {
+            Optional<DatasetMetadata.DataStore> dataStore = getDataStore(sink);
+            dataStore.ifPresentOrElse(
+                    extractLineageMetadata(records),
+                    () -> LOGGER.warn("Current {} sink is not supported to emit OpenLineage output dataset", sink));
+        }
+
+    }
+
+    private Consumer<DatasetMetadata.DataStore> extractLineageMetadata(List<ChangeEvent<Object, Object>> records) {
+        return store -> records.forEach(record -> {
+            SourceRecord sourceRecord = ((EmbeddedEngineChangeEvent<Object, Object, Object>) record).sourceRecord();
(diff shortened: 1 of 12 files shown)


debezium-server-core/src/main/java/io/debezium/server/BaseChangeConsumer.java
```diff
@@ -106,4 +125,33 @@ protected Map<String, String> convertHeaders(ChangeEvent<Object, Object> record)
...
+        catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+        records.forEach(record-> {
+            ConnectHeaders headers = new ConnectHeaders();
+
+            convertHeaders(record).forEach(headers::addString);
+            ConnectorContext connectorContext = ConnectorContext.from(headers);
+            String theName = "the name"; //TODO resolve based on the sink type
```
jpechane: I wonder if we can solve this metadata via annotations on the sink.

debezium-server-core/src/main/java/io/debezium/server/BaseChangeConsumer.java
```diff
@@ -106,4 +125,33 @@ protected Map<String, String> convertHeaders(ChangeEvent<Object, Object> record)
...
+            DatasetMetadata.DataStore dataStore = KAFKA; //TODO revolve based on the sink type
+
+            List<DatasetMetadata.FieldDefinition> fieldDefinitions = datasetDataExtractor.extract(((EmbeddedEngineChangeEvent<Object, Object, Object>) record).sourceRecord());
+            DebeziumOpenLineageEmitter.emit(connectorContext, DebeziumTaskState.RUNNING,
+                    List.of(new DatasetMetadata(theName, OUTPUT, STREAM_DATASET_TYPE, dataStore, fieldDefinitions)));
+        });
+    }
+
+    public abstract void consumeBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer) throws Exception;
```
jpechane: @kmos With rewrite to Quarkus Extension, do you think this can be solved declarativelly via decorator instead of template method design pattern?

kmos: I didn't have time to provide some code to show but yes I thought the same. I think that is addressable with via decoration but I am not 100 %. Something like:

```java

@Capturing
@Openlineage
public void consume(ChangeEvent<SourceRecord> event) {
 // do something amazing
}
```

mfvitale: Well event now the `clean` approach is to let the `BaseChangeConsumer` be a concreate class that contains the common logic and takes as parameter the specific sink implementation. In this way the `DebeziumServer` must adopt to this. 

I didn't proposed since in the end we will re-write DS with the extension.

We can decide to add the support now and then re-work a bit or just post-pone when we have the extension based DS. @jpechane  WDYT?

jpechane: @mfvitale I personally would prefer a change that would not affect sink source code even if it would be a less clean and then do the "proper" solution with rewrite to the Quarkus.

mfvitale: I think the proposed code changes are the minimal we can do to achieve the result with the current implementation. Do you have any other idea?

jpechane: How about adding a decorator around https://github.com/debezium/debezium-server/blob/main/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java#L140 ?

mfvitale: @jpechane Sounds good. I modified the code with the decorator. If we agree on it I can finalize the implementation for the OpenLineage stuff.

debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
```diff
@@ -0,0 +1,76 @@
...
+    public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+
+        final String sink = config.getValue("debezium.sink.type", String.class);
+
+        try {
+            this.delegateConsumer.handleBatch(records, committer);
+        }
+        catch (Exception e) {
+            throw new RuntimeException(e);
```
jpechane:
```suggestion
            throw new DebeziumException(e);
```

debezium-server-core/src/main/java/io/debezium/server/BaseChangeConsumer.java
jpechane: Yes, that's nice!

debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        records.forEach(record -> {
```
vjuranek: Should we always enable it? IMHO there should be at least option to turn it off, ideally turn off by default and enabled only when user request it.

mfvitale: by default a No-Op lineage emitter is used.

vjuranek: But we still do this additional data processing, which is not needed at all and slows down the things. I believe we should skip whole this loop if the lineage is not turned on.

mfvitale: Yes, I can do the optimization.

debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
```diff
@@ -0,0 +1,103 @@
...
+        final String sink = config.getValue(PROP_SINK_TYPE, String.class);
+
+        try {
+            this.delegateConsumer.handleBatch(records, committer);
+        }
+        catch (Exception e) {
+            throw new DebeziumException("Error while executing batch", e);
+        }
+
+        Boolean isOpenLineageEnabled = config.getOptionalValue(PROP_SOURCE_PREFIX + OPEN_LINEAGE_INTEGRATION_ENABLED, Boolean.class).orElse(false);
```
jpechane: This could be calculated only once in constructor. Also in that case I recommend to unbox it.

debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
```diff
@@ -0,0 +1,103 @@
...
+        if (isOpenLineageEnabled) {
+            Optional<DatasetMetadata.DataStore> dataStore = getDataStore(sink);
```
jpechane, line 65: Now the sink type and OL datastore are tighly coupled. Would it be possible to provide a config option where you could override the datastore? So for example my sink is PubSub but OL datatore would be Kafka?

mfvitale: Well this should not be arbitrary. It important for the dataset `name` and `namespace` to be as  defined by [OpenLineage spec](https://openlineage.io/docs/spec/naming/) so that producers/consumers can match the same. 

As you can see a lot of our sink has no official convention. So I decided to add the support only for Kafka and if then someone request we can go through adding it and propose a convention also to OpenLineage specification. WDYT?

jpechane: Thanks for the clarification!

debezium-server-core/src/main/java/io/debezium/server/DefaultChangeConsumer.java
mfvitale: > Also in that case I recommend to unbox it.

What do you mean?

jpechane: Using `boolean` instead of `Boolean`.

vjuranek (approved):
Nice, LGTM now.

debezium-server-core/src/test/java/io/debezium/server/DebeziumServerTest.java
```diff
@@ -36,7 +36,7 @@ public class DebeziumServerTest {
 
     private static final int MESSAGE_COUNT = 5;
 
-    {
+    static {
```
jpechane: What is the motivation for the change? I'd argue that you want to clear offsets before each test.

jpechane (approved):
@mfvitale LGTM, thanks! I left one question, when addressed the PR can be merged.

debezium-server-core/src/test/java/io/debezium/server/DebeziumServerTest.java
mfvitale: Reverted, It was correct as before.

PR #257, opened 2026-03-16
DBZ#1668 Add descriptors generation to Debezium Server sinks

### LINKED ISSUE #1668: Add descriptors generation to Debezium Server sinks
Also Debezium Server sink descriptor must be generated

mfvitale:
closes: debezium/dbz#1668

requires: https://github.com/debezium/debezium/pull/7274

### DIFF
--- a/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
+++ b/debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
@@ -35,7 +35,7 @@ import io.debezium.metadata.ComponentMetadata;
 import io.debezium.metadata.ComponentMetadataFactory;
 import io.debezium.server.BaseChangeConsumer;
 import io.debezium.server.CustomConsumerBuilder;
-import io.debezium.server.DebeziumServerSink;
+import io.debezium.server.api.DebeziumServerSink;
 
 /**
  * This sink adapter delivers change event messages to Azure Event Hubs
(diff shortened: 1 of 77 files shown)


debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
```diff
@@ -83,26 +80,25 @@ void connect() {
...
+        // Load configuration
+        io.debezium.config.Configuration configuration = io.debezium.config.Configuration.from(getConfigSubset(mpConfig, PROP_PREFIX));
+        this.config = new EventHubsChangeConsumerConfig(configuration);
 
-        // optional config
-        maxBatchSize = config.getOptionalValue(PROP_MAX_BATCH_SIZE, Integer.class).orElse(0);
-        configuredPartitionId = config.getOptionalValue(PROP_PARTITION_ID, String.class).orElse("");
-        configuredPartitionKey = config.getOptionalValue(PROP_PARTITION_KEY, String.class).orElse("");
+        configuredPartitionId = (config.getConfiguredPartitionId() != null) ? config.getConfiguredPartitionId() : "";
+        configuredPartitionKey = (config.getConfiguredPartitionKey() != null) ? config.getConfiguredPartitionKey() : "";
```
Naros: Could we use `Strings.defaultIfEmpty` for these types of cases?

debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
```diff
@@ -65,17 +67,20 @@ void connect() {
...
-        final Optional<String> password = config.getOptionalValue(PASSWORD, String.class);
+        final Config mpConfig = ConfigProvider.getConfig();
+
+        // Load configuration
+        io.debezium.config.Configuration configuration = io.debezium.config.Configuration.from(getConfigSubset(mpConfig, CONF_PREFIX));
+        this.config = new InfinispanSinkConsumerConfig(configuration);
+
+        final String serverHost = config.getServerHost();
+        final String cacheName = config.getCacheName();
+        final Integer serverPort = config.getServerPort() != null ? config.getServerPort() : ConfigurationProperties.DEFAULT_HOTROD_PORT;
```
Naros: For situations like this, would it make sense to abstract this detail away inside the various sink-specific configuration class so for this call site it would simply be `config.getServerPort()` ? We could probably extend this to a variety of other places and minimize redundancy.

Naros (commented):
Hi @mfvitale I left just a few inline questions.

debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
jpechane: Or a helper method so it will be like Util.defaultWhenNull(config.getServerPort(),ConfigurationProperties.DEFAULT_HOTROD_PORT)

kmos (approved):
Apart from what we discussed offline, my only suggestion is (if possible) to move the interfaces into another package, such as `*.api` or something similar.

debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
mfvitale: Well since we use the `io.debezium.config.Configuration` inside the configuration classes we can use it to get the defaults and or properly set default on `Field`

mfvitale:
@Naros @jpechane Could you take another pass?

debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
jpechane: @mfvitale Good point! This should be added in all places but I am fine doing that kind of cleanup in a separate PR.

jpechane (approved):

PR #269, opened 2026-04-14
debezium/dbz#1386 Fix infinite loop when batch contains only heartbeat messages in Redis sink

Binayak490-cyber:
## Problem

When `skip.heartbeat.messages=true` and a batch contains only heartbeat records, `handleBatch()` enters an infinite loop. Heartbeat records were marked processed via `committer.markProcessed()` but never removed from `clonedBatch` because `processedRecords` was declared after the heartbeat-skip block. This kept `clonedBatch.size()` above 0 forever, so `completedSuccessfully` never became `true`. The `continue` in the heartbeat block also bypassed `delayStrategy.sleepWhen()`, making it a tight spin with no sleep.

## Fix

Moved `processedRecords` declaration to before the heartbeat-skip block, added heartbeat records to it, and called `clonedBatch.removeAll(processedRecords)` right after the for-loop. If `clonedBatch` is empty at that point, `completedSuccessfully` is set to `true` and the loop exits cleanly.

## Testing

Added a unit test in `RedisMemoryThresholdTest` that calls `handleBatch()` with a heartbeat-only batch using a mocked `RedisClient` (no Docker needed). The test confirmed the infinite loop before the fix and passes after.

### DIFF
--- a/debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
+++ b/debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
@@ -5,14 +5,25 @@
  */
 package io.debezium.server.redis;
 
+import java.lang.reflect.Field;
 import java.util.AbstractMap.SimpleEntry;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 import io.debezium.config.Configuration;
+import io.debezium.engine.ChangeEvent;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.engine.DebeziumEngine.RecordCommitter;
+import io.debezium.engine.Header;
 import io.debezium.storage.redis.RedisClient;
 import io.debezium.util.Collect;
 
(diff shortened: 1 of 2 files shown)


Binayak490-cyber:
Here is the issue - https://github.com/debezium/dbz/issues/1386

Hi @kmos, @Naros, @jpechane, as discussed I've opened this PR to fix the infinite loop in `RedisStreamChangeConsumer.handleBatch()` when a batch contains only heartbeat messages.

The root cause was that `processedRecords` was declared after the heartbeat-skip block, so heartbeat records were never removed from `clonedBatch`, keeping `completedSuccessfully` false forever. The fix moves `processedRecords` before the block and tracks heartbeat records in it so they get properly removed.

Also included a unit test in `RedisMemoryThresholdTest` that reproduces the infinite loop without needing a running Redis instance. Need your review and feedback!

Naros (approved):

debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
```diff
@@ -24,6 +35,57 @@ public class RedisMemoryThresholdTest {
...
+        setField(consumer, RedisStreamChangeConsumer.class, "heartbeatPrefix", HEARTBEAT_PREFIX);
+        setField(consumer, RedisStreamChangeConsumer.class, "client", new RedisClientImpl(_10MB, _20MB));
+        setField(consumer, RedisStreamChangeConsumer.class, "redisMemoryThreshold",
+                new RedisMemoryThreshold(new RedisClientImpl(_10MB, _20MB), consumerConfig));
+
+        List<ChangeEvent<Object, Object>> batch = List.of(new HeartbeatChangeEvent(HEARTBEAT_PREFIX + ".testc"));
+        RecordCommitter<ChangeEvent<Object, Object>> committer = new NoOpRecordCommitter();
+
+        ExecutorService executor = Executors.newSingleThreadExecutor();
+        Future<?> future = executor.submit(() -> {
```
kmos: try to refactor lines from 60 to 80 with `Awailatility`: [example](https://stackoverflow.com/questions/71346616/how-can-i-await-at-least-specified-amount-of-time-with-awaitility#72032661)

Binayak490-cyber: Thanks @kmos for the resource and now it is done, refactored the full block (lines 60-80) to use Awaitility -replaced ExecutorService, Future, and TimeoutException with AtomicBoolean + Awaitility.await().atMost(3, TimeUnit.SECONDS).untilAsserted() as suggested. Thanks for the reference!

debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
```diff
@@ -24,6 +33,51 @@ public class RedisMemoryThresholdTest {
...
+            try {
+                consumer.handleBatch(batch, committer);
+                completed.set(true);
+            }
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+        }).start();
+
+        Awaitility.await()
```
kmos: something like:

```java
public class RedisMemoryThresholdTest {
	// [...]
    private final ExecutorService executor = Executors.newSingleThreadExecutor();
	// [...]

    @Test
    public void testHeartbeatOnlyBatchCompletesWithoutInfiniteLoop() throws Exception {
    // [...]
	        Awaitility.await()
                .atMost(3, TimeUnit.SECONDS)
                .until(consume(consumer, batch, committer)::isDone);
	}
	
    private Future<?> consume(RedisStreamChangeConsumer consumer,
                            List<ChangeEvent<Object, Object>> changeEvents,
                            RecordCommitter<ChangeEvent<Object, Object>> committer) {
      return executor.submit(() -> {
                  try {
                      consumer.handleBatch(changeEvents, committer);
                  }
                  catch (InterruptedException e) {
                      throw new RuntimeException(e);
                  }
              });
  }
}
```

Binayak490-cyber: @kmos, Done, refactored as suggested - added ExecutorService as a class field, extracted the submit logic into a private consume() helper method, and updated the test to use Awaitility.await().atMost(3, TimeUnit.SECONDS).until(consume(consumer, batch, committer)::isDone). Thanks for the guidance! Can you check it out now @kmos, that is this the one which you were telling me to change it to...?

kmos (approved):
LGTM!

Binayak490-cyber:
Thanks @kmos, @Naros!

PR #271, opened 2026-04-21
debezium/dbz#1840 Apache Fluss Debezium Server sink

### LINKED ISSUE #1840: Introduce an incubating Apache Fluss Debezium Sink
[Apache Fluss](https://fluss.apache.org/) is a streaming storage system designed for real-time analytics and AI workloads. It sits at the real-time data layer of Lakehouse architectures, serving as the bridge between raw streaming data and the broader data lakehouse, handling low-latency ingestion and high-throughput processing while integrating with engines such as Apache Flink and Apache Spark.

Today, getting data from a source database like PostgreSQL into Fluss typically means going through one of two paths:

```
PostgreSQL => Debezium => Kafka => Flink => Analytics/Lakehouse
```
```
PostgreSQL => Flink CDC/Debezium => Flink => Analytics/Lakehouse
```

Both work, but native Debezium support for Fluss would open up some meaningful improvements. The core idea is a clean separation of concerns with a Kafka-less, lightweight footprint model, in which capture, storage, and compute each become independently deployable, scalable, and operable.

The pipeline would become
```
PostgreSQL => Debezium Server => Fluss => Flink => Analytics/Lakehouse
```
Or leveraging Fluss' built-in lakehouse tiering:
```
PostgreSQL => Debezium Server => Fluss => Lakehouse (Paimon/Iceberg)
```

This would particularly shine if users want to maintain a single CDC pipeline for both real-time and batch analytics. Such systems might include real-time fraud detection while also supplying the same event stream to analytics for trend/report analysis.

Naros:
Fixes debezium/dbz#1840

- [x] Support all Debezium's native logical types
- [x] Documentation


vjuranek (approved):
LGTM, great work @Naros !

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
```diff
@@ -0,0 +1,357 @@
...
+            // Extracted new record state — whole struct is the row
+            op = Envelope.Operation.CREATE.code();
+            after = valueStruct;
+            before = null;
+            rowSchema = valueSchema;
+        }
+
+        final RowType rowType = descriptor.getSchema().getRowType();
+
+        if (hasPrimaryKey) {
```
mfvitale: Could maybe add a configuration lo also let users choose if they want to use primary-key tables or log tables?

mfvitale:
@Naros Great work!  I left a comment.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
```diff
@@ -0,0 +1,357 @@
...
+
+    private static final String PROP_PREFIX = "debezium.sink.fluss.";
+    private static final Duration RETRY_INTERVAL = Duration.ofSeconds(1);
+
+    private final ComponentMetadataFactory componentMetadataFactory = new ComponentMetadataFactory();
+    private final Map<TablePath, AppendWriter> appendWriters = new ConcurrentHashMap<>();
+    private final Map<TablePath, UpsertWriter> upsertWriters = new ConcurrentHashMap<>();
+    private final Map<TablePath, TableDescriptor> tableDescriptorCache = new ConcurrentHashMap<>();
+
+    FlussChangeConsumerConfig config;
```
jpechane, line 80: Why are not the fields private?

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
Naros: @mfvitale, would you agree making it an enum where the user can pick between `auto`, `upsert`, or `append` mode?

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
```diff
@@ -0,0 +1,357 @@
...
+    protected SourceRecord toSourceRecord(ChangeEvent<Object, Object> record) {
+        return ((EmbeddedEngineChangeEvent) record).sourceRecord();
+    }
+
+    private void writeRecordWithRetry(ChangeEvent<Object, Object> record, TablePath tablePath,
+                                      TableDescriptor descriptor, boolean hasPrimaryKey)
+            throws InterruptedException {
+
+        int attempts = 0;
+        while (true) {
```
jpechane: I beleieve we have similar code elswhere. I think it would be good to either add a method to `BaseChangeConsumer`  or add a class an `Awaitlity`-like class to our util library that will do the repeat, counting, delays etc.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Converter for all Kafka Connect to Apache Fluss type mapping.
+ *
+ * <p>Handles both schema-level conversion (Connect {@link org.apache.kafka.connect.data.Schema} to Fluss
+ * {@link DataType}) and runtime value conversion (Connect field values to Fluss row values).
+ *
+ * @author Chris Cranford
+ */
+public class FlussTypeConverter {
```
jpechane, line 68: Please convert to non-static class. I can imagine in future we might have conversion modified base on the configuration or any other state.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
```diff
@@ -0,0 +1,269 @@
...
+
+    private static final long MILLIS_PER_DAY = 86_400_000L;
```
jpechane: Please see `io.debezium.time.Conversions`

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * Converts a Kafka Connect {@link org.apache.kafka.connect.data.Schema} to the matching Fluss {@link DataType}.
+     *
+     * @param connectSchema the Connect field schema
+     * @return corresponding Fluss DataType
+     */
+    static DataType toFlussDataType(org.apache.kafka.connect.data.Schema connectSchema) {
+        final String schemaName = Strings.defaultIfBlank(connectSchema.name(), "");
+
+        if (LOGICAL_NAME.equals(schemaName)) {
```
jpechane: I wonder if we can use `case ... when ...` for `switch` expression to put it everyhing into one?

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
```diff
@@ -0,0 +1,269 @@
...
+     * @return the value in the form expected by Fluss {@link org.apache.fluss.row.GenericRow#setField}
+     */
+    static Object toFlussValue(Object value, org.apache.kafka.connect.data.Schema schema) {
+        if (value == null || schema == null) {
+            return null;
+        }
+
+        final String schemaName = Strings.defaultIfBlank(schema.name(), "");
+
+        if (LOGICAL_NAME.equals(schemaName)) {
```
jpechane: Same proposal as above

debezium-server-fluss/src/test/java/io/debezium/server/fluss/FlussTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Manages the lifecycle of a Fluss cluster (ZooKeeper + coordinator + tablet server) for integration testing.
+ *
+ * <p>Uses host networking so that the Fluss coordinator and tablet server register with addresses
+ * reachable by the test JVM without port-mapping complications.
+ */
+public class FlussTestResourceLifecycleManager implements QuarkusTestResourceLifecycleManager {
+
+    private static final String FLUSS_IMAGE = "apache/fluss:0.9.0-incubating";
```
jpechane: Should go into `io.debezium.server.Images` ?

debezium-server-fluss/src/test/java/io/debezium/server/fluss/FlussTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,136 @@
...
+            zookeeper.start();
+            coordinator.start();
+            tablet.start();
+            createTestTable();
+            running.set(true);
+        }
+    }
+
+    // Pre-create the customers table as a log table (no primary key) so the test
+    // does not depend on the auto-create feature or schema-enabled JSON format.
```
jpechane: But I think we need also test to verify auto-creation feature

jpechane (commented):
@Naros Nice job, thanks! I left a couple of comments. Please take a look if it makes sense.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
mfvitale: Totally I agree. I was thinking the same but I didn't want to influence you :D

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
Naros: They're package private so that `FlussUnitTest` can mock them. I can rework that if you'd prefer for these to be private rather than package private.

jpechane: Please just use `@VisibleForTesting` or add a comment

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
Naros: Thanks, I hadn't seen the `RetryExecutor` added, will use it.  Thanks :rocket:

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
Naros: It's definitely doable, but it does make me question the readability. I'll add it in a separate commit and if we decide against it due to readability, we can drop it.

debezium-server-fluss/src/test/java/io/debezium/server/fluss/FlussTestResourceLifecycleManager.java
Naros: Added tests for auto-create, connect format, and for Mario's new upsert mode. That should give us coverage across the test suite for all those flavors I believe.

debezium-server-fluss/src/test/java/io/debezium/server/fluss/FlussTestResourceLifecycleManager.java
Naros: done

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
Naros: Fixed

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussTypeConverter.java
Naros: Done

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
Naros: Added, let me know what you think.

mfvitale: I have a question. If I get it, to store table you need to have primaryKeys but could I still decide to store in append mode even with primaryKeys? In opposite if you do not have primaryKeys you should not able to store on table but only on append log?

jpechane (approved):
@Naros Nice job, thanks! WRT the `switch` approach I like it but let's hear at least one more opinion.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
vjuranek: Current implementation doesn't allow to store table with primary key into log table in Fluss AFAICT. IMHO we should allow users to do so if they want to and just ignore that there is a primary key and consider it as an ordinary column.

mfvitale: Yes, this is exactly what I was trying to convey. If just was not sure if this is an option or not. So basically if the user want store in the table we need to have the primary key, if they just want to append the events in the log table ignoring the primary key as it was a kafka topic.

Naros: So to be sure I understand, if the primary key mode is `append` and the table has a primary key, we want to explicitly ignore that so that we can effectively use append-only mode in a keyless table.

Naros:
@jpechane Should we take silence as we have a quorum?

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
vjuranek: @Naros do you intend to add this option, so that users can switch to `APPEND` mode even when primery key exists?

vjuranek (commented):
I think we have still pending discussion about `APPEND` mode.

Naros:
Thanks for the reminder @vjuranek, I must have missed that. I'll take a look at that later today.

debezium-server-fluss/src/main/java/io/debezium/server/fluss/FlussChangeConsumer.java
Naros: @vjuranek added in the latest commit.

vjuranek (approved):
Thanks @Naros , no objections or anything else to discuss from my side.

vjuranek:
Anyone else any comments or it can be merged?

mfvitale (approved):

vjuranek:
I think we have a quorum now :-)

vjuranek:
Merged, nice work @Naros !

PR #272, opened 2026-04-22
debezium/dbz#1822 upgrade google libraries-bom to 26.74.0

nathan-smit-1:
Hi there,

This PR is to upgrade the google libraries-bom to the most recent version.  From my testing, this is the most recent version of libraries-bom that we can upgrade to without also upgrading protobuf to protobuf 4 (whose dependency is currently inherited from the Debezium parent).

I've done some testing on my side and all looks to be working fine.


vjuranek (approved):

vjuranek:
Merged, thanks @nathan-smit-1 !

PR #279, opened 2026-05-20
align Quarkus version for Operator, Debezium Server Platform and Extensions in Debezium to 3.33.1.1 debezium/dbz#1951

kmos:
closes debezium/dbz#1951

depends on https://github.com/debezium/debezium/pull/7463

### DIFF
--- a/debezium-server-core/pom.xml
+++ b/debezium-server-core/pom.xml
@@ -130,6 +130,7 @@
         <dependency>
             <groupId>io.apicurio</groupId>
             <artifactId>apicurio-registry-utils-converter</artifactId>
+            <version>${version.apicurio}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/debezium-server-milvus/pom.xml
+++ b/debezium-server-milvus/pom.xml
@@ -76,7 +76,6 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers-milvus</artifactId>
             <scope>test</scope>
-            <version>1.21.3</version>
         </dependency>
         <dependency>
             <groupId>io.debezium</groupId>
--- a/debezium-server-milvus/pom.xml
+++ b/debezium-server-milvus/pom.xml
@@ -76,6 +76,7 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers-milvus</artifactId>
             <scope>test</scope>
+            <version>1.21.3</version>
         </dependency>
         <dependency>
             <groupId>io.debezium</groupId>
--- a/debezium-server-pubsub/pom.xml
+++ b/debezium-server-pubsub/pom.xml
@@ -93,6 +93,7 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers-gcloud</artifactId>
             <scope>test</scope>
+            <version>1.21.3</version>
         </dependency>
     </dependencies>
     <build>
--- a/debezium-server-pubsub/pom.xml
+++ b/debezium-server-pubsub/pom.xml
@@ -93,7 +93,6 @@
             <groupId>org.testcontainers</groupId>
             <artifactId>testcontainers-gcloud</artifactId>
             <scope>test</scope>
-            <version>1.21.3</version>
         </dependency>
     </dependencies>
     <build>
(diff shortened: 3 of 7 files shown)


kmos:
The pipeline is broken due to apicurio version. It should be updated as well to be compatible with Quarkus `3.33.1.1` in the debezium parent/bom

mfvitale:
Seems https://github.com/debezium/debezium-server/pull/280 trying to solve the same.

vjuranek:
> Seems #280 trying to solve the same.

not completely the same, IMHO we should include specific testcontainers into the depedencies. It was missing, probably because it was included by transitive dependecy, and this just reveals the issue. And lastly, it unblock the build, which IMHO should be fixed ASAP.

kmos:
> > Seems #280 trying to solve the same.
> 
> not completely the same, IMHO we should include specific testcontainers into the depedencies. It was missing, probably because it was included by transitive dependecy, and this just reveals the issue. And lastly, it unblock the build, which IMHO should be fixed ASAP.

@vjuranek do you mean milvus, qdrant and gcloud ? They are included in this PR + fix for redis. What is missed is the upgrade to Apicurio. I have tested the entire build against Apicurio `3.2.4`

kmos:
> > Seems #280 trying to solve the same.
> 
> not completely the same, IMHO we should include specific testcontainers into the depedencies. It was missing, probably because it was included by transitive dependecy, and this just reveals the issue. And lastly, it unblock the build, which IMHO should be fixed ASAP.

@vjuranek I just have seen your PR and it seems that they changed some artifactID. I don't which should be the right name for the containers.

mfvitale:
> > > Seems #280 trying to solve the same.
> > 
> > 
> > not completely the same, IMHO we should include specific testcontainers into the depedencies. It was missing, probably because it was included by transitive dependecy, and this just reveals the issue. And lastly, it unblock the build, which IMHO should be fixed ASAP.
> 
> @vjuranek I just have seen your PR and it seems that they changed some artifactID. I don't which should be the right name for the containers.

Starting from version 2.x they changes the names from to `org.testcontainers:x` to org.testcontainers:testcontainers-x`. When we updated to [version 2.x](https://github.com/debezium/dbz/issues/1615) we reflected the new naming on some containers but, as @vjuranek said, maybe there was some transitive dependencies that bring the old version. So I suppose the fix should be to use the new naming. WDYT?

kmos:
> > > > Seems #280 trying to solve the same.
> > > 
> > > 
> > > not completely the same, IMHO we should include specific testcontainers into the depedencies. It was missing, probably because it was included by transitive dependecy, and this just reveals the issue. And lastly, it unblock the build, which IMHO should be fixed ASAP.
> > 
> > 
> > @vjuranek I just have seen your PR and it seems that they changed some artifactID. I don't which should be the right name for the containers.
> 
> Starting from version 2.x they changes the names from to `org.testcontainers:x` to org.testcontainers:testcontainers-x`. When we updated to [version 2.x](https://github.com/debezium/dbz/issues/1615) we reflected the new naming on some containers but, as @vjuranek said, maybe there was some transitive dependencies that bring the old version. So I suppose the fix should be to use the new naming. WDYT?

for me it's perfectly fine. The important thing is that together with the Quarkus upgrade, it's necessary to fix the redis test. We can merge the @vjuranek PR and go head with this one for the redis test

kmos:
@mfvitale the important thing is that the actual failing is caused by Apicurio version which I have tested locally

mfvitale:
> @mfvitale the important thing is that the actual failing is caused by Apicurio version which I have tested locally

Yes, apicurio is another movie :D

mfvitale (approved):

debezium-server-bom/pom.xml
```diff
@@ -91,6 +91,12 @@
                 <scope>import</scope>
             </dependency>
 
+            <dependency>
+                <groupId>org.hibernate.orm</groupId>
+                <artifactId>hibernate-agroal</artifactId>
+                <version>${version.hibernate}</version>
+            </dependency>
```
Naros: With the change in the main repo to add this to the Debezium BOM, is this needed here in the BOM?

Naros (commented):
Overall LGTM, just one question below.

debezium-server-bom/pom.xml
kmos: you are totally right! 🤖

PR #280, opened 2026-05-20
debezium/dbz#1962 Add missing testcontainers dependencies

vjuranek:
Fixes https://github.com/debezium/dbz/issues/1962

### DIFF
--- a/debezium-server-milvus/pom.xml
+++ b/debezium-server-milvus/pom.xml
@@ -74,7 +74,7 @@
         </dependency>
         <dependency>
             <groupId>org.testcontainers</groupId>
-            <artifactId>milvus</artifactId>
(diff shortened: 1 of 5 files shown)


vjuranek:
There seems to be another issue (test-related) with Apicurio version:-(

vjuranek:
As mentioned under #279 , IMHO we should include the missing dependencies anyway and unblock the build. Apicurio issue can be fixed in separate PR (#279)

mfvitale:
@vjuranek I agree this, apart from apicurio, is something that should be fixed anyway.

kmos (approved):

kmos:
my suggestion is to merge this pr and then proceed to merge the #279 after apicurio override. In this way we have a green build

PR #288, opened 2026-07-04
debezium/dbz#2078 Guard sink-to-storage property remapping against explicit storage config

mahitha-ada:
Fixes debezium/dbz#2078

## Problem
Debezium Server reuses the `debezium.sink.<type>.*` connection properties for the schema history and offset storage namespaces. The remapping runs unconditionally and is only guarded per-key by `overwrite=false`, so when the sink and storage use different property names for the same concept the sink values can silently override explicit storage configuration, and unrelated sink properties leak into the storage namespaces.

## Fix
Skip the sink-to-storage remapping for a namespace once it has been configured explicitly; the convenience behavior is unchanged when no explicit storage configuration is present. The logic now lives in `DebeziumServerConfigSourceFactory` (moved from `DebeziumServer.java` in debezium/dbz#1731).

## Testing
Added a unit test for `DebeziumServerConfigSourceFactory` covering the reuse and skip paths (no Docker); it fails before the fix and passes after.

### DIFF
--- a/debezium-server-core/src/main/java/io/debezium/server/configuration/DebeziumServerConfigSourceFactory.java
+++ b/debezium-server-core/src/main/java/io/debezium/server/configuration/DebeziumServerConfigSourceFactory.java
@@ -84,12 +84,24 @@ public class DebeziumServerConfigSourceFactory implements ConfigSourceFactory {
         ConfigValue sink = context.getValue(PROP_SINK_TYPE);
         if (sink != null && sink.getValue() != null) {
             remapped.put(QUARKUS_DEBEZIUM_PREFIX + "name", sink.getValue());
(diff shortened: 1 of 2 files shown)


mahitha-ada:
Rebased this PR on latest `main` to pick up `debezium/dbz#1174` / `a99ed0be`, which fixes the JDBC API mismatch that caused the previous Cross Maven CI failure.

The old failing run was on commit `c40e38bc`; the PR head is now `3c939186`. The new Cross Maven CI run for this head is currently marked `action_required`, so it looks like it needs maintainer approval before it can run.

Naros (approved):
This generally LGTM, lets just wait on CI to see if things pass.

Naros:
Thanks @mahitha-ada, applied. :tada:

PR #291, opened 2026-07-08
debezium/dbz#1174 Align with core changes

mfvitale:
Fix compilation issue introduced by https://github.com/debezium/debezium/pull/7576


jpechane:
@mfvitale Applied, thanks!

PR #292, opened 2026-07-14
debezium/dbz#2225 Support MessageGroupId on standard SNS topics for SQS fair queues

### LINKED ISSUE #2225: SNS/SQS sinks: support MessageGroupId on standard topics/queues for AWS fair queues
## Feature request or enhancement

**Which use case/requirement will be addressed by the proposed feature?**

Allow the SNS and SQS sinks to set `MessageGroupId` on **standard** topics/queues (not just FIFO), to support [AWS SQS fair queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-fair-queues.html).

Since 2025-07-31, `MessageGroupId` on a **standard** queue is used as a *tenant identifier* for noisy-neighbor mitigation (it does not enforce ordering there — that stays FIFO-only). SNS standard topics [forward `MessageGroupId`](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html) to subscribed standard SQS queues to activate it automatically.

Use case: a CDC → SQS pipeline fanning changes into a multi-tenant **standard** queue wants to set a per-tenant group id (e.g. from a record field/header) and get fairness for free — no consumer changes.

Today this isn't config-reachable. The SNS sink only sets the group id when the topic ARN ends in `.fifo` ([`SnsChangeConsumer.java`](https://github.com/debezium/debezium-server/blob/main/debezium-server-sns/src/main/java/io/debezium/server/sns/SnsChangeConsumer.java)):

```java
boolean isFifo = topicArn.endsWith(".fifo");
if (isFifo) {
    String groupId = headers.getOrDefault(config.getMessageGroupIdHeader(), null);
    if (groupId == null) groupId = rawKey != null ? asString(rawKey) : config.getFifoDefaultGroupId();
    builder.messageGroupId(groupId);
}
```

The SQS sink is analogous (its group-id config is FIFO-oriented), so driving fair queues from Debezium currently requires patching the sink or a custom bean.

(Same ecosystem catch-up is tracked in [LocalStack #13008](https://github.com/localstack/localstack/issues/13008) and [terraform-provider-aws #44063](https://github.com/hashicorp/terraform-provider-aws/issues/44063).)

---

**Implementation ideas (optional)**

Let `MessageGroupId` be set on standard topics/queues too, opt-in, sourced from a record header (falling back to the record key), regardless of topic/queue type:

- **Preferred:** a type-agnostic `debezium.sink.{sns,sqs}.message.group.id.header` (+ optional default), applied whenever configured; keep the existing `fifo.*` properties as back-compat aliases.
- **Minimal alternative:** un-gate the existing header-based group-id logic from the `.fifo` check.

Set via `PublishRequest.messageGroupId(...)` (SNS) and `SendMessageRequest` / `SendMessageBatchRequestEntry.messageGroupId(...)` (SQS).

**Backward compatibility:** fully compatible — FIFO behavior unchanged; the group id is only set on standard topics/queues when the new/ungated config is explicitly provided (`MessageGroupId` is optional on standard queues, so no-config = today's behavior).

Happy to submit the PR (both sinks + a test + docs note) — would appreciate a steer on the preferred config shape first.

yongwang-eng:
## What

Adds an opt-in `debezium.sink.sns.message.group.id.enabled` (default `false`) that lets the SNS sink set `MessageGroupId` on **standard** (non-FIFO) topics, enabling [Amazon SQS fair queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-fair-queues.html) on subscribed standard queues.

## Why

Since 2025, `MessageGroupId` on a **standard** SQS queue acts as a *tenant identifier* for fair-queue noisy-neighbor mitigation (it does **not** enforce ordering there — that stays FIFO-only), and [SNS standard topics forward `MessageGroupId`](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html) to subscribed standard SQS queues. Today the SNS sink only sets `MessageGroupId` when the topic ARN ends in `.fifo`, so a CDC → SNS → standard-SQS pipeline can't get per-tenant fairness. This closes that gap.

## Change

- New config `message.group.id.enabled` (boolean, default `false`).
- When `true`, `MessageGroupId` is resolved for standard topics the **same way as FIFO** — from the group-id header, falling back to the record key, then the configured default.
- **Neutral config namespace (per review):** the group-id resolution properties are now `message.group.id.header` and `message.group.id.default`. The former `fifo.message.group.id.header` and `fifo.default.group.id` are registered as **deprecated aliases** (`Field.withDeprecatedAliases`) — existing configs keep working, with a deprecation warning logged when an alias key is used and the new key takes precedence when both are set. `fifo.message.dedup.id.header` is intentionally unchanged, as `MessageDeduplicationId` is FIFO-only (invalid on standard topics).
- **FIFO topics are unaffected** (they always set `MessageGroupId`, as before).
- Default `false` preserves existing behavior for all current standard-topic users.

## Tests

Behavior (SnsUnitTest):
- standard topic + flag on → resolves `MessageGroupId` from the header,
- standard topic + flag on → falls back to the record key when no header,
- standard topic + flag on → falls back to `message.group.id.default` when no header and no key,
- standard topic + flag **off** (default) → sets **no** `MessageGroupId` (backward compatibility),
- **FIFO regression** → with the flag on, a FIFO topic still sets **both** `MessageGroupId` and `MessageDeduplicationId`.

Config resolution (SnsChangeConsumerConfigTest):
- new keys resolve; deprecated `fifo.*` aliases still resolve,
- new key wins when both old and new are set (both properties),
- defaults hold (`aggregateId` / `default`), and the flag defaults to `false`.

## Notes

- **Behavior note:** on a standard topic with the flag on, a record with no header and no key resolves to `message.group.id.default` (`"default"`), i.e. one shared tenant bucket. That's opt-in-by-design (provide a key/header for real per-tenant fairness); documented in the config description.
- **Docs:** a config-reference PR in `debezium/debezium` will follow — I'll link it here.
- **Scope:** SNS sink only. Extending the SQS *direct* sink to a dynamic per-message group id is a separately scoped follow-up, per review discussion.

Closes debezium/dbz#2225

### DIFF
--- a/debezium-server-sns/src/main/java/io/debezium/server/sns/SnsChangeConsumer.java
+++ b/debezium-server-sns/src/main/java/io/debezium/server/sns/SnsChangeConsumer.java
@@ -218,15 +218,20 @@ public class SnsChangeConsumer extends BaseChangeConsumer implements DebeziumSer
             builder.messageAttributes(attributes);
         }
 
-        if (isFifo) {
-            // MessageGroupId: use header value if present, fallback to raw event key, then default
+        // MessageGroupId: set for FIFO topics (message ordering) or, when explicitly enabled, for standard
+        // topics as well (Amazon SQS fair-queue tenant identifier — no ordering on standard topics).
+        if (isFifo || config.isMessageGroupIdEnabled()) {
+            // use header value if present, fallback to raw event key, then default
             String groupId = headers.getOrDefault(config.getMessageGroupIdHeader(), null);
             if (groupId == null) {
                 Object rawKey = event.record().key();
-                groupId = rawKey != null ? asString(rawKey) : config.getFifoDefaultGroupId();
+                groupId = rawKey != null ? asString(rawKey) : config.getDefaultMessageGroupId();
             }
             builder.messageGroupId(groupId);
+        }
 
+        // MessageDeduplicationId applies only to FIFO topics
+        if (isFifo) {
             // MessageDeduplicationId: use header if configured and present
             if (config.getMessageDeduplicationIdHeader() != null) {
                 String dedupId = headers.get(config.getMessageDeduplicationIdHeader());
@@ -315,9 +320,10 @@ public class SnsChangeConsumer extends BaseChangeConsumer implements DebeziumSer
                 SnsChangeConsumerConfig.TOPIC_ARN,
                 SnsChangeConsumerConfig.TOPIC_ARN_PREFIX,
                 SnsChangeConsumerConfig.DEFAULT_RETRIES,
-                SnsChangeConsumerConfig.FIFO_MESSAGE_GROUP_ID_HEADER,
-                SnsChangeConsumerConfig.FIFO_MESSAGE_DEDUP_ID_HEADER,
-                SnsChangeConsumerConfig.FIFO_DEFAULT_GROUP_ID);
+                SnsChangeConsumerConfig.MESSAGE_GROUP_ID_HEADER,
+                SnsChangeConsumerConfig.MESSAGE_GROUP_ID_DEFAULT,
+                SnsChangeConsumerConfig.MESSAGE_GROUP_ID_ENABLED,
+                SnsChangeConsumerConfig.FIFO_MESSAGE_DEDUP_ID_HEADER);
     }
 
     @Override
--- a/debezium-server-sns/src/main/java/io/debezium/server/sns/SnsChangeConsumerConfig.java
+++ b/debezium-server-sns/src/main/java/io/debezium/server/sns/SnsChangeConsumerConfig.java
@@ -6,6 +6,8 @@
 package io.debezium.server.sns;
 
 import org.apache.kafka.common.config.ConfigDef;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.debezium.config.Configuration;
 import io.debezium.config.Field;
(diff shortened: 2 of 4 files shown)


Naros (approved):

Naros:
Hi @yongwang-eng, the code LGTM, just waiting on CI and then it can be merged.

Naros:
Hi @yongwang-eng WRT to the open questions:

> Config naming: I reused the existing fifo.message.group.id.header / fifo.default.group.id for resolution to keep the change minimal (rather than proliferate config). Happy to introduce neutral message.group.id.* names (with fifo.* aliases) if you'd prefer a cleaner shape — let me know.

I think having a neutral `message.group.id.*` namespace makes far more sense, and then if we need to retain the FIFO ones for backward compatibility, those could be registered as deprecated alias keys.

> happy to follow up with a config-reference entry for message.group.id.enabled in the debezium/debezium docs if you'd like it in this PR or a separate one.

Yes, please raise a docs PR in the main repo for these changes, if you haven't already.

> SNS sink only. The SQS direct sink currently sets a static, FIFO-gated group id; extending it to a dynamic per-message group id for fair queues is a larger change — glad to follow up separately.

I definitely agree, SQS can be handled in a separately scoped PR.

yongwang-eng:
Thanks @Naros — done. The group-id properties now use the neutral namespace: message.group.id.header and message.group.id.default, with fifo.message.group.id.header / fifo.default.group.id registered as deprecated aliases (via Field.withDeprecatedAliases; the new key wins when both are set, and a deprecation warning is logged when an alias key is used). I left fifo.message.dedup.id.header as-is since MessageDeduplicationId is FIFO-only — happy to move that one too if you'd prefer. Added alias-resolution and precedence tests alongside the existing ones, and updated the PR description to match.

Naros:
Hi @yongwang-eng, could we add one final test for the standard topic that exercises the fallback to the default message group ID? I believe Test 11 tests this with FIFO, but there isn't a non-FIFO equivalent.

yongwang-eng:
Good catch @Naros — added.

Naros:
Thanks @yongwang-eng, applied :tada:

PR #293, opened 2026-07-22
debezium/dbz#2265 Enable support for all JdbcSinkConnector database dialects in debezium-server-jdbc

rk3rn3r:
+ add MySQL tests

closes debezium/dbz#2265


debezium-server-core/src/test/java/io/debezium/server/Images.java
```diff
@@ -34,4 +34,5 @@ public class Images {
     public static final String FLUSS_IMAGE = "apache/fluss:0.9.0-incubating";
     public static final String FLUSS_ZOOKEEPER_IMAGE = "zookeeper:3.9";
     public static final String POSTGRES_IMAGE = "mirror.gcr.io/library/postgres:16-alpine";
+    public static final String MYSQL_IMAGE = "container-registry.oracle.com/mysql/community-server:8.0";
```
Naros:
```suggestion
    public static final String MYSQL_IMAGE = "container-registry.oracle.com/mysql/community-server:8.2";
```
In the core repo, the lowest we support right now is 8.2.

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
```diff
@@ -220,14 +123,21 @@ public void close() {
...
     }
+
+    private static Map<String, String> toStringMap(Map<String, Object> source) {
+        final Map<String, String> result = new HashMap<>();
+        source.forEach((key, value) -> {
+            if (value != null) {
+                result.put(key, value.toString());
+            }
+        });
+        return result;
```
Naros:
```suggestion
        return source.entrySet().stream()
            .filter(e -> e.getValue() != null)
            .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().toString()));
```
What do you think about using modern streams?

debezium-server-jdbc/src/test/java/io/debezium/server/jdbc/JdbcChangeConsumerMySqlIT.java
```diff
@@ -0,0 +1,33 @@
...
+ * - Schema evolution
+ */
+@QuarkusTest
+@TestProfile(JdbcMySqlTestProfile.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public class JdbcChangeConsumerMySqlIT extends JdbcChangeConsumerIT {
+
+    @Override
+    protected JdbcTestUtils testUtils() {
+        return new JdbcMySqlTestUtils();
```
Naros: While `JdbcMySqlTestUtils` and the other dialect equivalents are generally stateless, this creates a new instance on each invocation of `testUtils()` within a test class, and as we can see in the `JdbcChangeConsumerIT`, this can be half a dozen or more times per method.

Could we instead treat this method as a lazy initializer that caches the instance when created, returning the cached copy if its non-null? I don't see why we need a new instance for every method call.

debezium-server-jdbc/pom.xml
```diff
@@ -24,6 +24,11 @@
             <artifactId>debezium-connector-jdbc</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-sink</artifactId>
+        </dependency>
```
Naros: Can't this dependency be dropped? It's already a transitive dependency of the JDBC sink connector artifact, right?

Naros (commented):
Hi @rk3rn3r a quick pass with some feedback.

debezium-server-jdbc/pom.xml
rk3rn3r: Ah, this is a leftover from a previous version that used SinkProgressionListener explicitly. This got removed with using the JdbcSinkTask directly and I think it can be removed now.

debezium-server-jdbc/src/test/java/io/debezium/server/jdbc/JdbcChangeConsumerMySqlIT.java
rk3rn3r: Yep, that's great. Will do

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
rk3rn3r: I love it.

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
```diff
@@ -18,116 +19,57 @@
...
-                    CONNECTOR_LOGICAL_NAME,
-                    CONNECTOR_NAME,
-                    TASK_ID,
-                    Module.version(),
-                    UUID.randomUUID(),
-                    new java.util.HashMap<>());
+            final Config mpConfig = ConfigProvider.getConfig();
+            final Map<String, String> props = toStringMap(getConfigSubset(mpConfig, PROP_PREFIX));
 
-            this.changeEventSink = new JdbcChangeEventSink(
```
mfvitale, line 123: Could I ask the reason of this change? The original idea was to leave this sink decoupled from KC things. What is the advantage of this?

debezium-server-core/src/test/java/io/debezium/server/Images.java
rk3rn3r: Done

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
rk3rn3r: Done

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
rk3rn3r: Where do you see more Kafka API deps than before?
We would need to duplicate almost the entire code inside `JdbcSinkConnectorTask` here and there is "only" a dependency problem that we even can't resolve until connectors are entirely refactored to be Kafka / KC dependency free, it's not even used anywhere except for `SinkRecord` (which was used in the previous version as well). Because of this the previous version was also only able to run Postgres with the UnnestRecordWriter because only that code path was moved here.
In my last 1:1 with @jpechane we also agreed on that the first iteration using the Kafka API is fine.

debezium-server-jdbc/src/test/java/io/debezium/server/jdbc/JdbcChangeConsumerMySqlIT.java
rk3rn3r: Done

rk3rn3r: Done

rk3rn3r:
@Naros I've addressed all comments.

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
mfvitale: > Where do you see more Kafka API deps than before?

`task` is a KC concept isn't it?

In any way. This can be solved when connectors will be KC dep free. 

mine here, was just a curiosity. Thanks for the explanation.

rk3rn3r: @mfvitale I filed https://github.com/debezium/dbz/issues/2300 for a K-deps-free version of the `JdbcSinkConnectorTask` for decoupling. wdyt?

mfvitale: Nice, thanks!

rk3rn3r:
@Naros Wdyt, can we merge this one?

debezium-server-jdbc/pom.xml
Naros: @rk3rn3r this seems to still be in the diff, could it not be removed?

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
```diff
@@ -136,82 +77,43 @@ void connect() {
...
-
-            for (BatchEvent record : events.records()) {
-                record.commit();
-            }
+        final Collection<SinkRecord> sinkRecords = events.records().stream()
+                .map(converter::convert)
+                .collect(Collectors.toList());
 
-            LOGGER.debug("Successfully processed batch of {} records", events.records().size());
+        task.put(sinkRecords);
```
Naros: There is a change in exception behavior here. If there is ever a possibility that `put` throws an exception that isn't stored as the "Last processing exception," then this propagates that exception up rather than being wrapped in a `DebeziumException` like before.

Should we consider at least wrapping this in a `try-block`, as I don't think throwing a `ConnectException` up to Debezium Server without it being wrapped makes logical sense.

rk3rn3r:
```
        catch (Throwable throwable) {

            // Capture failure
            LOGGER.error("Failed to process record: {}", throwable.getMessage(), throwable);
            previousPutException = throwable;

            // Stash all records
            records.forEach(this::markNotProcessed);
        }
```

Agreed, the highly critical path is definitely secured but the less critical one of setting offsets is not. Will add an additional `try/catch` block.

debezium-server-jdbc/pom.xml
rk3rn3r: @Naros it's only in the debezium-server bom, where it potentially makes sense, do you think I should remove it? I thought it would be better to keep it if the dependency graph somehow includes it now or later. wdyt, still removing it?

Naros: The reason I suggested removing it is that it sets a precedent that hasn't been present, adding extra Debezium-specific transitive dependencies that are always brought in by the submodules themselves. It's not harmful, but it's just not something we've consistently done here.

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/JdbcChangeConsumer.java
rk3rn3r: Done

rk3rn3r:
The Debezium Server Redis Sink Adapter test error is unrelated.

Naros:
Applied, thanks @rk3rn3r

PR #294, opened 2026-07-27
debezium/dbz#2279 Add Databricks Zerobus Ingest sink for Debezium Server

### LINKED ISSUE #2279: Add Databricks Zerobus Ingest sink adapter to Debezium Server
## Feature request or enhancement

**Which use case/requirement will be addressed by the proposed feature?**

Databricks has launched **Zerobus Ingest**, now generally available on AWS and Azure. This serverless ingestion API allows users to write records directly into Unity Catalog-managed Delta tables, removing the need for a message broker.
Zerobus is ideal for change data capture (CDC). Currently, sending Debezium change events to Delta requires setting up Kafka and a Spark/Connect consumer. With a native Debezium Server sink for Zerobus, the process simplifies to: `source DB → Debezium Server → Delta`, fulfilling the need for direct CDC integration into the lakehouse.

Zerobus exposes three client interfaces we can target:
- **gRPC** via the official `com.databricks:zerobus-ingest-sdk` (GA) — the primary path;
- **REST** (`POST <endpoint>/zerobus/v1/tables/<catalog.schema.table>/insert`) — good for
  environments without a persistent connection;
- a **Kafka-compatible producer endpoint** — which the existing `kafka` sink can
  target once we supply an OAUTHBEARER login handler for Databricks OAuth.

Relevant Zerobus semantics that shape the sink design:
- at-least-once delivery (consumers dedupe downstream);
- one stream binds to one managed Delta table (so topic→table routing maps 1:1);
- strict schema validation, JSON objects only (tombstones must be skipped);
- OAuth via a service principal, using an RFC 9396 `authorization_details` token scoped to
  the `zerobusDirectWriteApi` audience.

**Would the maintainers welcome this as an in-tree sink** (alongside `kinesis`, `pubsub`,
`fluss`, etc.), or would you prefer it live as an external/community module? Happy to align
on scope and structure before opening a PR.

---

**Implementation ideas (optional)**

I have a working prototype (built against 3.6.0.Final, to be ported to `main`/3.7 for the
PR) structured like the existing sink modules:

- New module `debezium-server-databricks-zerobus` with a `ChangeConsumer` per route:
  - `zerobus` — gRPC via the Databricks SDK (`createJsonStream` / `ingestRecordOffset` /
    `flush`), one `ZerobusJsonStream` cached per target table;
  - `zerobusrest` — REST insert per record;
  - plus a `ZerobusOAuthBearerLoginCallbackHandler` so the native `kafka` sink can
    authenticate against the Zerobus Kafka endpoint (Beta).
- Flat JSON output (`ExtractNewRecordState`, `schemas.enable=false`), with tombstones
  skipped since Zerobus only accepts JSON objects.
- Validated end-to-end (Postgres CDC → managed Delta): snapshot, insert, update, and delete
  (`delete.tombstone.handling.mode=rewrite`, `__deleted=true`), across multiple tables with
  regex topic→table routing.

Open questions I'd like maintainer input on before the PR:
1. **Integration tests** — other sinks use Testcontainers, but Zerobus is a proprietary
   cloud service with no local emulator. How would you prefer CI coverage (e.g., unit tests
   for encoding/auth/routing + IT gated behind credentials, similar to how cloud sinks are
   handled)?
2. **SDK dependency** — is a dependency on `com.databricks:zerobus-ingest-sdk` (Apache-2.0)
   acceptable in-tree, or should the gRPC path be optional?
3. **Scope of the first PR** — land the gRPC route alone first, then follow up with REST and
   the Kafka OAUTHBEARER handler in separate issues/PRs?

I'm happy to do the work and maintain the module.

### LINKED ISSUE #2378: Add a native Databricks ZeroBus sink with acknowledgment-safe offset handling
## Feature request or enhancement

**Which use case/requirement will be addressed by the proposed feature?**

Debezium Server users need a brokerless path for writing change events directly to Databricks ZeroBus Ingest and Unity Catalog managed tables. The contribution should make the delivery and source-offset boundary explicit: records must be written and durably acknowledged before Debezium marks them processed.

I would like to contribute a focused native Debezium Server sink named `zerobus`, limited to the official Databricks ZeroBus Java SDK path. The first contribution would not add REST or Kafka-compatible routes. It would provide a stable raw CDC envelope for JSON and Protobuf, preserving source-position metadata for ordering and downstream idempotency.

Would the maintainers welcome this focused module in the Debezium Server repository? I would appreciate guidance on the first-PR scope, SDK/JNI packaging in the distribution, and the preferred long-term integration-test strategy before opening a PR.

---

**Implementation ideas (optional)**

A working implementation currently includes:

- A new `debezium-server-zerobus` module selected with `debezium.sink.type=zerobus`.
- JSON and Protobuf streams using a stable envelope containing target table, destination, partition, operation, idempotency key, key, value, source position, and headers.
- Source, explicit, and regex table-routing modes for pre-created Unity Catalog managed tables.
- Source-order-preserving batching across interleaved target tables.
- Write and durable acknowledgement before Debezium offset commit.
- At-least-once semantics with deterministic source-position idempotency keys.
- Separate retry ownership: outer retries only for confirmed pre-delivery failures, with no full-chunk replay after an ambiguous ingest or acknowledgement outcome.
- Tombstone handling, native event filtering, encoded-record size limits, bounded LRU stream caching, and JMX delivery/safety metrics.
- Mocked unit tests for configuration, routing, filtering, serialization, batching, acknowledgements, retries, offset compatibility, metrics, and SDK packaging.
- A credential-gated live integration test that remains outside the default test run.

The module uses `com.databricks:zerobus-ingest-sdk:1.3.0`. The proposed PR would include module documentation and Debezium Server BOM/distribution wiring after agreement on scope.

wagnercsantos:
Adds a new Debezium Server sink that writes change events directly into **Databricks Zerobus Ingest** → managed Delta tables in Unity Catalog, with no intermediate broker.
Fixes debezium/dbz#2279
Fixes debezium/dbz#2378

## What

The new module (`debezium-server-databricks-zerobus`) ships three delivery routes and a schema-driven type mapping:

- **`zerobus`** — gRPC via the Databricks Zerobus Java SDK (GA transport, primary route);
- **`zerobusrest`** — Zerobus REST API, one HTTP request per record (no persistent connection);
- **Kafka OAUTHBEARER login handler** — lets the built-in `kafka` sink authenticate against the Zerobus Kafka-compatible endpoint;
- **`ZerobusTypeSystem`** — maps Connect/Debezium logical types to the Delta column contract (decimal→exact string, temporals→epoch, bytes→base64, JSON/JSONB→VARIANT, arrays→ARRAY&lt;T&gt;), the per-dialect "type system" equivalent for this sink; plus a small `StringifyFields` SMT for the VARIANT-per-field case.

Delivery is at-least-once; the sink carries the source LSN/offset for downstream dedup.

## Note on scope

When @Naros and I discussed this on debezium/dbz#2279, the plan was to land the gRPC route first and follow up with REST and the Kafka handler separately. Everything ended up ready in time, so this PR brings the **complete sink** as one coherent unit. **Happy to split it into the phased gRPC-first sequence if you'd prefer to review it that way** — the routes are cleanly separated and easy to peel apart; I brought them together only because the sink tells a more complete story whole.

## SDK dependency

The gRPC route depends on `com.databricks:zerobus-ingest-sdk`, a JNI wrapper over a native client that ships per-platform native libraries. It is declared **`<optional>true</optional>`**, so it is **not** propagated into the default Debezium Server distribution — the module compiles and unit-tests against it, but deployments opt in by adding the jar for the gRPC route. The `zerobusrest` route and the type system need nothing from the SDK.

## How it was validated

End-to-end against managed Delta tables, exercising snapshot + insert/update/delete:

- **Sources:** Postgres, MySQL, SQL Server, Db2, MongoDB, and Oracle (LogMiner, self-managed).
- **Scale:** a sustained run of ~2.5M change events with no ingestion errors.
- **Data types:** decimals (incl. thousands/locale edge cases → exact `DECIMAL`), temporals (DATE/TIMESTAMP/TIMESTAMP_NTZ, tz→string), binary/base64→BINARY, arrays→ARRAY&lt;T&gt;, JSON/JSONB→VARIANT, CLOB/BLOB.
- **Regional safety:** decimal serialization is locale-independent (verified with the JVM booted in a comma-decimal locale).

## Tests

59 tests, green (`mvn -pl debezium-server-databricks-zerobus test`):

- **Unit (no cloud):** JSON/type encoding (`ZerobusTypeSystem`), OAuth `authorization_details` (RFC 9396 — scoped grants, never `ALL_PRIVILEGES`), topic→table routing / qualified-name handling, tombstone-skip, and the `StringifyFields` SMT.
- **REST route:** mocked `HttpClient` asserting request path, headers and JSON body, plus tombstone-skip and HTTP-error handling.
- **gRPC route:** the `ZerobusJsonStream` is mocked with Mockito (mirroring how the Zerobus SDK tests itself), covering ingest+flush, skip, and fail-fast; the live gRPC IT is gated behind `@EnabledIfEnvironmentVariable` so it is skipped in the default CI run and executed manually with credentials.

## Docs

The operations-guide section for this sink is submitted separately, against the docs in the main repo — debezium/debezium#7725 (config properties, managed-table prerequisites + RFC 9396 grants, table routing, and the data type mapping table). A module-level `application.properties` example is included in this PR.

## Related platform fixes (separate PRs, required to run this sink with some sources)

Building/validating this sink surfaced a few pre-existing platform bugs that block certain sources on 3.7; each is contributed separately and linked here for context:

- debezium/dbz#2284 — Quarkus engine/Agroal bootstrap: MongoDB startup failure **and** Db2 silent no-op. Fix: debezium/debezium-quarkus#71.
- debezium/dbz#2304 — Db2 LOB serialization. Fix: debezium/debezium-connector-db2#237.
- debezium/dbz#2305 — Oracle archive-dest validation on `initial_only`. Fix: debezium/debezium#7724.

cc @Naros

### DIFF
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
@@ -14,8 +14,6 @@ import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 
-import io.debezium.server.databricks.zerobus.ZerobusOAuthBearerLoginCallbackHandler.TokenHttpClient;
-
 /**
  * Default {@link TokenHttpClient} backed by the JDK {@link HttpClient}. Posts an
  * {@code application/x-www-form-urlencoded} body to the Databricks OIDC token endpoint.
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server.databricks.zerobus;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+
+import io.debezium.server.databricks.zerobus.ZerobusOAuthBearerLoginCallbackHandler.TokenHttpClient;
+
+/**
+ * Default {@link TokenHttpClient} backed by the JDK {@link HttpClient}. Posts an
+ * {@code application/x-www-form-urlencoded} body to the Databricks OIDC token endpoint.
+ */
+class DefaultTokenHttpClient implements TokenHttpClient {
+
+    private static final Duration CONNECT_TIMEOUT = Duration.ofSeconds(10);
+    private static final Duration REQUEST_TIMEOUT = Duration.ofSeconds(30);
+
+    private final HttpClient httpClient = HttpClient.newBuilder()
+            .connectTimeout(CONNECT_TIMEOUT)
+            .build();
+
+    @Override
+    public String postForm(String url, String authorizationHeader, String formBody) throws IOException {
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(URI.create(url))
+                .timeout(REQUEST_TIMEOUT)
+                .header("Authorization", authorizationHeader)
+                .header("Content-Type", "application/x-www-form-urlencoded")
+                .header("Accept", "application/json")
+                .POST(HttpRequest.BodyPublishers.ofString(formBody, StandardCharsets.UTF_8))
+                .build();
+
+        try {
+            HttpResponse<InputStream> response = httpClient.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            String body = new String(response.body().readAllBytes(), StandardCharsets.UTF_8);
+            if (response.statusCode() < 200 || response.statusCode() >= 300) {
+                throw new IOException("Zerobus token endpoint returned HTTP " + response.statusCode() + ": " + body);
+            }
+            return body;
+        }
+        catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IOException("Interrupted while requesting Zerobus OAuth token", e);
+        }
+    }
+}
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/StringifyFields.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/StringifyFields.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server.databricks.zerobus;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.json.JsonConverter;
+import org.apache.kafka.connect.storage.Converter;
+import org.apache.kafka.connect.transforms.Transformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Kafka Connect SMT that serializes selected record-value fields to JSON strings.
+ * <p>
+ * Databricks Zerobus only accepts a JSON <em>string</em> for a {@code VARIANT} column; it rejects a
+ * native struct/array/map ("invalid type: map/sequence, expected a string"). When flattening a
+ * document (e.g. the MongoDB {@code ExtractNewDocumentState} SMT), nested fields arrive as native
+ * structs/arrays, which cannot be written to a VARIANT column directly.
+ * <p>
+ * This SMT converts the configured fields into their JSON-string representation, so scalar fields
+ * can stay strongly typed while a single schema-flexible field (e.g. {@code properties}) is written
+ * to a VARIANT column. Fields that are already strings, or absent, are left untouched.
+ * <p>
+ * Configuration: {@code fields} — comma-separated list of top-level value field names to stringify.
+ * <p>
+ * Error behavior: if a targeted field cannot be serialized to JSON, the SMT logs the offending
+ * record's context (field name, schema type/name, value class and value) at ERROR level and
+ * re-throws a {@link ConnectException} (fail-fast). The stream halts and the offset is not
+ * committed past the record, so nothing is dropped silently; set {@code errors.tolerance=all} on
+ * the engine if you prefer to skip such records instead.
+ */
+public class StringifyFields<R extends ConnectRecord<R>> implements Transformation<R> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StringifyFields.class);
+
+    public static final String FIELDS_CONFIG = "fields";
+
+    private static final ConfigDef CONFIG_DEF = new ConfigDef().define(
+            FIELDS_CONFIG, ConfigDef.Type.LIST, ConfigDef.NO_DEFAULT_VALUE, ConfigDef.Importance.HIGH,
+            "Comma-separated list of value field names to serialize to a JSON string.");
+
+    private Set<String> fields;
+    private final JsonConverter jsonConverter = new JsonConverter();
+
+    @Override
+    public void configure(Map<String, ?> configs) {
+        // Debezium/Kafka Connect passes SMT config values as raw strings (not parsed
+        // through ConfigDef), so accept both a List and a comma-separated String.
+        Object raw = configs.get(FIELDS_CONFIG);
+        this.fields = new LinkedHashSet<>();
+        if (raw instanceof List) {
+            for (Object f : (List<?>) raw) {
+                this.fields.add(f.toString().trim());
+            }
+        }
+        else if (raw != null) {
+            for (String f : raw.toString().split(",")) {
+                if (!f.trim().isEmpty()) {
+                    this.fields.add(f.trim());
+                }
+            }
+        }
+        // schemas.enable=false → emit plain JSON without the Connect schema envelope.
+        Map<String, Object> jc = new HashMap<>();
+        jc.put("schemas.enable", false);
+        jc.put("converter.type", "value");
+        jsonConverter.configure(jc);
+    }
+
+    @Override
+    public R apply(R record) {
+        if (record.value() == null || !(record.value() instanceof Struct)) {
+            return record;
+        }
+        Struct value = (Struct) record.value();
+        Schema schema = value.schema();
+
+        // Build a new schema where the targeted fields become STRING.
+        SchemaBuilder builder = SchemaBuilder.struct().name(schema.name()).version(schema.version());
+        for (Field field : schema.fields()) {
+            if (fields.contains(field.name())) {
+                builder.field(field.name(), field.schema().isOptional() ? Schema.OPTIONAL_STRING_SCHEMA : Schema.STRING_SCHEMA);
+            }
+            else {
+                builder.field(field.name(), field.schema());
+            }
+        }
+        Schema newSchema = builder.build();
+
+        Struct newValue = new Struct(newSchema);
+        for (Field field : schema.fields()) {
+            Object fieldValue = value.get(field);
+            if (fields.contains(field.name()) && fieldValue != null && !(fieldValue instanceof String)) {
+                newValue.put(field.name(), toJsonString(field.name(), field.schema(), fieldValue));
+            }
+            else {
+                newValue.put(field.name(), fieldValue);
+            }
+        }
+
+        return record.newRecord(record.topic(), record.kafkaPartition(), record.keySchema(), record.key(),
+                newSchema, newValue, record.timestamp(), record.headers());
+    }
+
+    /** Serialize a single field value to its JSON representation using the Connect JsonConverter. */
+    private String toJsonString(String fieldName, Schema fieldSchema, Object fieldValue) {
+        try {
+            byte[] json = ((Converter) jsonConverter).fromConnectData("_stringify", fieldSchema, fieldValue);
+            return new String(json, java.nio.charset.StandardCharsets.UTF_8);
+        }
+        catch (Exception e) {
+            // Fail-fast: a field that cannot be serialized must stop the stream rather than be
+            // dropped or corrupted silently. Log enough context to diagnose the offending record
+            // (field name, schema type/name and value class), then re-throw so the engine halts
+            // and the offset is not committed past this record.
+            String schemaType = fieldSchema == null ? "null" : String.valueOf(fieldSchema.type());
+            String schemaName = fieldSchema == null ? "null" : String.valueOf(fieldSchema.name());
+            String valueClass = fieldValue == null ? "null" : fieldValue.getClass().getName();
+            LOGGER.error("StringifyFields failed to serialize field '{}' (schema type={}, schema name={}, "
+                    + "value class={}) to a JSON string. The record is not modified and the stream will halt "
+                    + "so no data is dropped or committed past this point. Value: {}",
+                    fieldName, schemaType, schemaName, valueClass, fieldValue, e);
+            throw new ConnectException("StringifyFields could not serialize field '" + fieldName
+                    + "' (schema type=" + schemaType + ", value class=" + valueClass + ") to a JSON string", e);
+        }
+    }
+
+    @Override
+    public ConfigDef config() {
+        return CONFIG_DEF;
+    }
+
+    @Override
+    public void close() {
+        jsonConverter.close();
+    }
+}
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
@@ -6,8 +6,10 @@
 package io.debezium.server.databricks.zerobus;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
@@ -102,6 +104,11 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
     @Override
     public void handle(CapturingEvents<BatchEvent> events) throws InterruptedException {
 
+        // Track only the streams that actually received a record in this batch, so we flush those
+        // rather than every open stream: with N tables but only a few touched per batch, flushing
+        // all N would add avoidable round-trips.
+        final Set<String> touchedTables = new HashSet<>();
+
         for (BatchEvent record : events.records()) {
             String json = toZerobusJson(record, this::getString);
             String table = resolveTable(record.destination());
@@ -109,6 +116,7 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
                 LOGGER.debug("Ingesting into {}: {}", table, json);
                 try {
                     stream(table).ingestRecordOffset(json);
+                    touchedTables.add(table);
                     metrics.recordIngested(operationOf(record), sourceTsMsOf(record));
                 }
                 catch (Exception e) {
@@ -126,15 +134,15 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
             record.commit();
         }
 
-        // Flush every touched stream to durability before acknowledging the batch (at-least-once).
+        // Flush only the streams touched in this batch to durability before acknowledging (at-least-once).
         final long flushStartNanos = System.nanoTime();
-        for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
+        for (String table : touchedTables) {
             try {
-                entry.getValue().flush();
+                streams.get(table).flush();
             }
             catch (Exception e) {
                 metrics.recordError();
-                throw new DebeziumException("Failed to flush Zerobus stream for table '" + entry.getKey() + "'", e);
+                throw new DebeziumException("Failed to flush Zerobus stream for table '" + table + "'", e);
             }
         }
         metrics.flushed((System.nanoTime() - flushStartNanos) / 1_000_000L);
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
@@ -131,10 +131,12 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
                 metrics.recordSkipped();
                 LOGGER.trace("Skipping record for destination '{}' (table={}): {}", record.destination(), table, json);
             }
-            record.commit();
         }
 
-        // Flush only the streams touched in this batch to durability before acknowledging (at-least-once).
+        // Flush only the streams touched in this batch to durability, then acknowledge the offsets.
+        // The commit must happen after the flush: committing first would advance the source offset
+        // while the ingested records are still buffered, so a crash between commit and flush would
+        // drop them (breaking the at-least-once guarantee).
         final long flushStartNanos = System.nanoTime();
         for (String table : touchedTables) {
             try {
@@ -146,6 +148,10 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
             }
         }
         metrics.flushed((System.nanoTime() - flushStartNanos) / 1_000_000L);
+
+        for (BatchEvent record : events.records()) {
+            record.commit();
+        }
         maybeLogMetrics();
     }
 
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
@@ -33,6 +33,7 @@ import io.debezium.runtime.CapturingEvents;
 import io.debezium.server.BaseChangeConsumer;
 import io.debezium.server.api.DebeziumServerConsumer;
 import io.debezium.server.api.DebeziumServerSink;
+import io.debezium.server.databricks.zerobus.metrics.ZerobusSinkMetrics;
 
 /**
  * Debezium Server sink that writes change events directly into Databricks Zerobus Ingest using the
@@ -61,6 +62,8 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
 
     private ZerobusChangeConsumerConfig config;
     private ZerobusSdk sdk;
+    private final ZerobusSinkMetrics metrics = new ZerobusSinkMetrics("grpc");
+    private long batchesSinceLog = 0;
 
     // One stream per target table (Zerobus: 1 stream = 1 table).
     private final Map<String, ZerobusJsonStream> streams = new HashMap<>();
@@ -72,6 +75,7 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
         this.config = new ZerobusChangeConsumerConfig(configuration);
 
         this.sdk = new ZerobusSdk(config.getEndpoint(), config.getWorkspaceUrl());
+        metrics.register();
         LOGGER.info("Zerobus gRPC sink connected: endpoint={}, workspaceUrl={}", config.getEndpoint(), config.getWorkspaceUrl());
     }
 
@@ -81,6 +85,7 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
         for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
             try {
                 entry.getValue().close();
+                metrics.streamClosed();
             }
             catch (Exception e) {
                 LOGGER.warn("Could not close Zerobus stream for table '{}'", entry.getKey(), e);
@@ -90,6 +95,7 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
         if (sdk != null) {
             sdk.close();
         }
+        metrics.unregister();
         LOGGER.info("Zerobus gRPC sink closed");
     }
 
@@ -103,8 +109,10 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
                 LOGGER.debug("Ingesting into {}: {}", table, json);
                 try {
                     stream(table).ingestRecordOffset(json);
+                    metrics.recordIngested(operationOf(record), sourceTsMsOf(record));
                 }
                 catch (Exception e) {
+                    metrics.recordError();
                     throw new DebeziumException("Failed to ingest record into Zerobus table '" + table + "'", e);
                 }
             }
@@ -112,19 +120,32 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
                 // Skip records that are not ingestable: tombstones / null payloads, and events whose
                 // destination is not a fully-qualified table (e.g. MySQL schema-history / DDL events,
                 // which the binlog connector emits on the topic-prefix "topic").
+                metrics.recordSkipped();
                 LOGGER.trace("Skipping record for destination '{}' (table={}): {}", record.destination(), table, json);
             }
             record.commit();
         }
 
         // Flush every touched stream to durability before acknowledging the batch (at-least-once).
+        final long flushStartNanos = System.nanoTime();
         for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
             try {
                 entry.getValue().flush();
             }
             catch (Exception e) {
+                metrics.recordError();
                 throw new DebeziumException("Failed to flush Zerobus stream for table '" + entry.getKey() + "'", e);
             }
+        }
+        metrics.flushed((System.nanoTime() - flushStartNanos) / 1_000_000L);
+        maybeLogMetrics();
+    }
+
+    private void maybeLogMetrics() {
+        final int interval = config.getMetricsLogInterval();
+        if (interval > 0 && ++batchesSinceLog >= interval) {
+            batchesSinceLog = 0;
+            metrics.logMetricsSummary();
         }
     }
 
@@ -180,6 +201,68 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
         return record.value() == null ? null : verbatim.apply(record.value());
     }
 
+    /**
+     * Extracts the Debezium operation ({@code c}/{@code u}/{@code d}/{@code r}) from a change event, or
+     * {@code null} when unavailable. Shared by the gRPC and REST routes to split ingestion metrics by
+     * operation. Handles both shapes the sink sees in practice:
+     * <ul>
+     * <li>the full Debezium envelope, which carries {@code op} at the top level;</li>
+     * <li>an unwrapped record (the recommended {@code ExtractNewRecordState} setup with
+     * {@code add.fields=op,...}), which exposes it as {@code __op} (default {@code __} prefix).</li>
+     * </ul>
+     */
+    static String operationOf(BatchEvent record) {
+        org.apache.kafka.connect.data.Struct struct = valueStruct(record);
+        if (struct == null) {
+            return null;
+        }
+        if (struct.schema().field("op") != null) {
+            return struct.getString("op");
+        }
+        if (struct.schema().field("__op") != null) {
+            return struct.getString("__op");
+        }
+        return null;
+    }
+
+    /**
+     * Extracts {@code source.ts_ms} (the source database change timestamp) from a change event, or
+     * {@code -1} when unavailable. Used to compute the sink's freshness / lag behind the source.
+     * Handles both the full envelope (nested {@code source.ts_ms}) and an unwrapped record with
+     * {@code add.fields=source.ts_ms} (exposed as the top-level {@code __source_ts_ms}).
+     */
+    static long sourceTsMsOf(BatchEvent record) {
+        org.apache.kafka.connect.data.Struct struct = valueStruct(record);
+        if (struct == null) {
+            return -1L;
+        }
+        // Envelope: nested source.ts_ms
+        if (struct.schema().field("source") != null && struct.get("source") instanceof org.apache.kafka.connect.data.Struct) {
+            org.apache.kafka.connect.data.Struct sourceStruct = (org.apache.kafka.connect.data.Struct) struct.get("source");
+            long ts = longField(sourceStruct, "ts_ms");
+            if (ts >= 0) {
+                return ts;
+            }
+        }
+        // Unwrapped: flattened __source_ts_ms
+        return longField(struct, "__source_ts_ms");
+    }
+
+    private static org.apache.kafka.connect.data.Struct valueStruct(BatchEvent record) {
+        Object value = record.record() == null ? null : record.record().value();
+        return value instanceof org.apache.kafka.connect.data.Struct ? (org.apache.kafka.connect.data.Struct) value : null;
+    }
+
+    private static long longField(org.apache.kafka.connect.data.Struct struct, String field) {
+        if (struct.schema().field(field) != null) {
+            Object v = struct.get(field);
+            if (v instanceof Number) {
+                return ((Number) v).longValue();
+            }
+        }
+        return -1L;
+    }
+
     /** Zerobus only accepts JSON objects; a null or "null"/non-object payload (a tombstone) must be skipped. */
     static boolean isJsonObject(String json) {
         if (json == null) {
@@ -199,7 +282,9 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
             StreamConfigurationOptions options = StreamConfigurationOptions.builder()
                     .setMaxInflightRecords(config.getMaxInflightRecords())
                     .build();
-            return sdk.createJsonStream(table, config.getClientId(), config.getClientSecret(), options).join();
+            ZerobusJsonStream openedStream = sdk.createJsonStream(table, config.getClientId(), config.getClientSecret(), options).join();
+            metrics.streamOpened();
+            return openedStream;
         }
         catch (Exception e) {
             throw new DebeziumException("Could not open Zerobus stream for table '" + table + "'", e);
@@ -214,7 +299,8 @@ public class ZerobusChangeConsumer extends BaseChangeConsumer
                 ZerobusChangeConsumerConfig.CLIENT_ID,
                 ZerobusChangeConsumerConfig.CLIENT_SECRET,
                 ZerobusChangeConsumerConfig.TABLE,
-                ZerobusChangeConsumerConfig.MAX_INFLIGHT_RECORDS);
+                ZerobusChangeConsumerConfig.MAX_INFLIGHT_RECORDS,
+                ZerobusChangeConsumerConfig.METRICS_LOG_INTERVAL);
     }
 
     @Override
--- a/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
+++ b/debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
@@ -1,0 +1,224 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server.databricks.zerobus;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.annotation.PreDestroy;
+import jakarta.enterprise.context.Dependent;
+import jakarta.inject.Named;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.databricks.zerobus.StreamConfigurationOptions;
+import com.databricks.zerobus.ZerobusJsonStream;
+import com.databricks.zerobus.ZerobusSdk;
+
+import io.debezium.DebeziumException;
+import io.debezium.Module;
+import io.debezium.config.Field;
+import io.debezium.metadata.ComponentMetadata;
+import io.debezium.metadata.ComponentMetadataFactory;
+import io.debezium.runtime.BatchEvent;
+import io.debezium.runtime.CapturingEvents;
+import io.debezium.server.BaseChangeConsumer;
+import io.debezium.server.api.DebeziumServerConsumer;
+import io.debezium.server.api.DebeziumServerSink;
+
+/**
+ * Debezium Server sink that writes change events directly into Databricks Zerobus Ingest using the
+ * Zerobus Java SDK (gRPC transport, GA). Change events are serialized as JSON by Debezium
+ * ({@code debezium.format.value=json}) and forwarded verbatim through a per-table
+ * {@link ZerobusJsonStream}.
+ * <p>
+ * Zerobus binds one stream to one managed Delta table. This consumer therefore keeps a stream per
+ * resolved table name. The target table for a record is the mapped destination (topic) name, unless
+ * a fully-qualified default table is configured via {@code debezium.sink.zerobus.table}.
+ * <p>
+ * Delivery is at-least-once: each batch is flushed to durability before the records are committed,
+ * so a crash after flush but before commit may re-send records. Deduplicate downstream using the
+ * Debezium source LSN/offset carried in the payload.
+ */
+@Named("zerobus")
+@Dependent
+public class ZerobusChangeConsumer extends BaseChangeConsumer
+        implements DebeziumServerConsumer<CapturingEvents<BatchEvent>>, DebeziumServerSink {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZerobusChangeConsumer.class);
+
+    private static final String PROP_PREFIX = "debezium.sink.zerobus.";
+
+    private final ComponentMetadataFactory componentMetadataFactory = new ComponentMetadataFactory();
+
+    private ZerobusChangeConsumerConfig config;
+    private ZerobusSdk sdk;
+
+    // One stream per target table (Zerobus: 1 stream = 1 table).
+    private final Map<String, ZerobusJsonStream> streams = new HashMap<>();
+
+    @PostConstruct
+    void connect() {
+        final Config mpConfig = ConfigProvider.getConfig();
+        io.debezium.config.Configuration configuration = io.debezium.config.Configuration.from(getConfigSubset(mpConfig, PROP_PREFIX));
+        this.config = new ZerobusChangeConsumerConfig(configuration);
+
+        this.sdk = new ZerobusSdk(config.getEndpoint(), config.getWorkspaceUrl());
+        LOGGER.info("Zerobus gRPC sink connected: endpoint={}, workspaceUrl={}", config.getEndpoint(), config.getWorkspaceUrl());
+    }
+
+    @PreDestroy
+    @Override
+    public void close() {
+        for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
+            try {
+                entry.getValue().close();
+            }
+            catch (Exception e) {
+                LOGGER.warn("Could not close Zerobus stream for table '{}'", entry.getKey(), e);
+            }
+        }
+        streams.clear();
+        if (sdk != null) {
+            sdk.close();
+        }
+        LOGGER.info("Zerobus gRPC sink closed");
+    }
+
+    @Override
+    public void handle(CapturingEvents<BatchEvent> events) throws InterruptedException {
+
+        for (BatchEvent record : events.records()) {
+            String json = toZerobusJson(record, this::getString);
+            String table = resolveTable(record.destination());
+            if (isJsonObject(json) && table != null) {
+                LOGGER.debug("Ingesting into {}: {}", table, json);
+                try {
+                    stream(table).ingestRecordOffset(json);
+                }
+                catch (Exception e) {
+                    throw new DebeziumException("Failed to ingest record into Zerobus table '" + table + "'", e);
+                }
+            }
+            else {
+                // Skip records that are not ingestable: tombstones / null payloads, and events whose
+                // destination is not a fully-qualified table (e.g. MySQL schema-history / DDL events,
+                // which the binlog connector emits on the topic-prefix "topic").
+                LOGGER.trace("Skipping record for destination '{}' (table={}): {}", record.destination(), table, json);
+            }
+            record.commit();
+        }
+
+        // Flush every touched stream to durability before acknowledging the batch (at-least-once).
+        for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
+            try {
+                entry.getValue().flush();
+            }
+            catch (Exception e) {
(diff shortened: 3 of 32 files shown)


wagnercsantos:
Cross-linking an architectural note for reviewers: @rk3rn3r's dbz#2300 (*"Split `JdbcSinkConnectorTask` to provide a version without Kafka API that can be integrated deps-free into Debezium Server"*) points in the same direction this sink already takes.

This sink is Kafka-API-free on the ingestion path by construction — it implements `DebeziumServerConsumer<CapturingEvents<BatchEvent>>` and consumes `BatchEvent`, rather than going through `SinkTask`/`SinkRecord`/`TopicPartition`. So there's no dependency between the two efforts and nothing blocks this PR.

The one point of overlap worth flagging: `ZerobusTypeSystem` intentionally reimplements, self-contained, the same responsibility the shared `io.debezium.sink.*` SPI (`DebeziumSinkRecord`, `FieldDescriptor`, `ColumnDescriptor`, `type/Type`, naming strategies) already provides for the JDBC sink — it doesn't import that SPI today. If dbz#2300 lands and that SPI becomes the canonical deps-free path for Debezium Server sinks, this sink is a natural candidate to migrate onto it in a follow-up. Happy to do that as a separate PR once the direction is settled. (Note both this type system and the shared SPI still rely on `org.apache.kafka.connect.data.*` `Schema`/`Struct` — "Kafka-API-free" here means free of the Kafka *runtime/client*, not the Connect data types, matching the `@TODO` already in `DebeziumSinkRecord`.)

wagnercsantos:
Follow-up: added **sink observability** to this PR (new commits on the branch), so the sink ships with the JMX metrics operators expect from a Debezium Server sink rather than logs alone.

## What

A JMX MBean per route, `debezium.zerobus:type=connector-metrics,context=sink,server=<grpc|rest>,task=0`, following the same convention as the JDBC sink connector and Debezium source connectors. Attributes:

- **Throughput:** `TotalRecordsIngested`, split by operation `TotalInserts` / `TotalUpdates` / `TotalDeletes` / `TotalReads` (`c`/`u`/`d`/`r`); `TotalRecordsSkipped` (tombstones / non-qualified destinations); `TotalErrors`; `TotalFlushes`.
- **Freshness:** `MilliSecondsBehindSource` = `now - source.ts_ms` of the last ingested event — the primary CDC-to-lakehouse signal (how far the Delta table trails the source).
- **Latency / state:** `LastFlushDurationMillis`, `MaxFlushDurationMillis`, `ActiveStreams`, `Route`.

An optional `debezium.sink.zerobus[rest].metrics.log.interval=N` also emits the same summary as an INFO line every N batches (default `0` = off; JMX is always on).

The metadata extractors handle both payload shapes the sink sees: the full Debezium envelope (`op`, nested `source.ts_ms`) and an unwrapped record (the recommended `ExtractNewRecordState` setup → flattened `__op` / `__source_ts_ms`). Missing fields degrade gracefully (no crash; only the per-op / freshness attributes stay empty).

Design note: the metrics are deliberately self-contained and do **not** implement the shared `io.debezium.sink.spi.SinkProgressListener` — mirroring how `ZerobusTypeSystem` owns its mapping rather than depending on `io.debezium.sink.*`. That keeps the sink free of the `debezium-sink` module dependency; if dbz#2300 lands and that SPI becomes the canonical deps-free contract, these are a natural candidate to migrate onto it in a follow-up.

## Tests & validation

- Unit test count is now **79** (was 59), green — new `ZerobusSinkMetricsTest` (counters, freshness math, flush min/max, JMX register/unregister) and `EnvelopeMetadataTest` (both envelope and unwrapped shapes), plus metrics-interaction assertions added to the existing gRPC/REST consumer tests.
- Validated live over remote JMX (a JConsole-style client): Postgres snapshot + streaming with `c`/`u`/`d`/`r` counters matching the applied DML, and `MilliSecondsBehindSource` moving; confirmed the same on MongoDB (which uses `ExtractNewDocumentState`).

Docs updated in the companion PR (debezium/debezium#7725) with a "Monitoring" section listing the attributes.

debezium-server-dist/pom.xml
```diff
@@ -356,12 +370,12 @@
...
+                    <artifactId>debezium-quarkus-mysql</artifactId>
                     <version>${project.version}</version>
                 </dependency>
                 <dependency>
                     <groupId>io.debezium.quarkus</groupId>
-                    <artifactId>debezium-quarkus-mysql</artifactId>
+                    <artifactId>debezium-quarkus-mariadb</artifactId>
                     <version>${project.version}</version>
                 </dependency>
                 <dependency>
```
Naros: These two changes, exchanging `mysql` and `mariadb`, seem superfluous. 
Could we drop those from the diff?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
```diff
@@ -0,0 +1,310 @@
...
+                LOGGER.trace("Skipping record for destination '{}' (table={}): {}", record.destination(), table, json);
+            }
+            record.commit();
+        }
+
+        // Flush every touched stream to durability before acknowledging the batch (at-least-once).
+        final long flushStartNanos = System.nanoTime();
+        for (Map.Entry<String, ZerobusJsonStream> entry : streams.entrySet()) {
+            try {
+                entry.getValue().flush();
```
Naros: Should we track exactly which streams were touched in this batch and only flush those? I'm curious: in a situation where you may have N streams, would calling N flushes rather than 1 or 2, if that's all that were touched in this batch, create unnecessary overhead costs?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusTokenProvider.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Base64;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Fetches and caches a Databricks OAuth token scoped to the Zerobus audience, using the
+ * client-credentials grant with RFC 9396 {@code authorization_details}. Shared by the REST route;
+ * the Kafka route has its own copy inside the SASL callback handler.
+ */
+class ZerobusTokenProvider {
```
Naros, line 15: The token exchange in this class and the BearerLoginCallback are identical; could that logic be pulled into a shared component to minimize code drift?

debezium-server-dist/pom.xml
```diff
@@ -379,6 +393,11 @@
                     <artifactId>debezium-quarkus-sqlserver</artifactId>
                     <version>${project.version}</version>
                 </dependency>
+                <dependency>
+                    <groupId>io.debezium.quarkus</groupId>
+                    <artifactId>debezium-quarkus-db2</artifactId>
+                    <version>${project.version}</version>
+                </dependency>
```
Naros: This looks like it may have been discovered as something missing? If so, could you raise this as a separate PR, as it may require backporting to 3.6?

Naros (commented):
Hi @wagnercsantos this looks quite good. I've left some comments.

debezium-server-dist/pom.xml
wagnercsantos: Good catch — those were an accidental reordering, not intentional. Reverted; the `pom.xml` diff is now limited to the Zerobus additions.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
wagnercsantos: Agreed — flushing every open stream is wasteful when a batch only touches a couple. Fixed: `handle()` now records the tables actually written in the batch and flushes only those. Added a test (`flushesOnlyTheStreamsTouchedInTheBatch`) asserting an untouched stream is not flushed. One intentional behavior change worth noting: a tombstone-only batch now flushes nothing (nothing was ingested), covered by the updated tombstone test.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusTokenProvider.java
wagnercsantos: Done. Pulled the exchange into a shared, stateless `ZerobusTokenExchange` (builds the `client_credentials` request with the RFC 9396 `authorization_details`, performs the call, and parses `access_token`/`expires_in`). Both routes now delegate to it — the REST `ZerobusTokenProvider` and the Kafka `ZerobusOAuthBearerLoginCallbackHandler` — so the request shape and parsing can't drift; each caller keeps only its own caching policy. Also promoted the small `TokenHttpClient` seam to a top-level type so both share it for testing.

debezium-server-dist/pom.xml
wagnercsantos: Right — that was a missing entry I ran into while validating the Db2 source, unrelated to this sink. I've dropped it from this PR and will raise it separately as you suggested, flagging it for a 3.6 backport.

wagnercsantos: Opened as debezium/debezium-server#296 (tracked by debezium/dbz#2318). Note: no 3.6 backport is needed — the `debezium-quarkus-db2` engine extension only exists on 3.7.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumerConfig.java
```diff
@@ -0,0 +1,117 @@
...
+    public static final Field CLIENT_ID = Field.create("client.id")
+            .withDisplayName("Service principal client id")
+            .withType(ConfigDef.Type.STRING)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withValidation(Field::isRequired)
+            .withDescription("OAuth client id (service principal application id).");
+
+    public static final Field CLIENT_SECRET = Field.create("client.secret")
+            .withDisplayName("Service principal client secret")
+            .withType(ConfigDef.Type.STRING)
```
Naros: We may want to define this as `ConfigDef.Type.PASSWORD`.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumerConfig.java
```diff
@@ -0,0 +1,114 @@
...
+    public static final Field CLIENT_ID = Field.create("client.id")
+            .withDisplayName("Service principal client id")
+            .withType(ConfigDef.Type.STRING)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withValidation(Field::isRequired)
+            .withDescription("OAuth client id (service principal application id).");
+
+    public static final Field CLIENT_SECRET = Field.create("client.secret")
+            .withDisplayName("Service principal client secret")
+            .withType(ConfigDef.Type.STRING)
```
Naros: We may want to define this as `ConfigDef.Type.PASSWORD` as well.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
```diff
@@ -0,0 +1,318 @@
...
+                }
+            }
+            else {
+                // Skip records that are not ingestable: tombstones / null payloads, and events whose
+                // destination is not a fully-qualified table (e.g. MySQL schema-history / DDL events,
+                // which the binlog connector emits on the topic-prefix "topic").
+                metrics.recordSkipped();
+                LOGGER.trace("Skipping record for destination '{}' (table={}): {}", record.destination(), table, json);
+            }
+            record.commit();
```
Naros, line 240: Should a record committal happen after the flush?

Naros (commented):
Hi @wagnercsantos, I found a few more items.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumerConfig.java
wagnercsantos: Done — changed `CLIENT_SECRET` to `ConfigDef.Type.PASSWORD`.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumerConfig.java
wagnercsantos: Done here as well.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
wagnercsantos: Good catch — yes. Moved the `record.commit()` calls to after the flush: the batch is ingested, the touched streams are flushed to durability, and only then are the offsets committed. Committing before the flush could advance the source offset while records were still buffered, so a crash in between would drop them and break the at-least-once guarantee. Added a test (`commitsOffsetsOnlyAfterTheStreamIsFlushed`) asserting the ingest -> flush -> commit ordering.

For the REST route this already held: `post()` is synchronous, so each record is durable (2xx) before its `commit()` — no separate flush step to reorder there.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@code application/x-www-form-urlencoded} body to the Databricks OIDC token endpoint.
+ */
+class DefaultTokenHttpClient implements TokenHttpClient {
+
+    private static final Duration CONNECT_TIMEOUT = Duration.ofSeconds(10);
+    private static final Duration REQUEST_TIMEOUT = Duration.ofSeconds(30);
+
+    private final HttpClient httpClient = HttpClient.newBuilder()
+            .connectTimeout(CONNECT_TIMEOUT)
+            .build();
```
Naros: In the REST code path, I believe we're creating 3 unique instances of `HttpClient`. Each one of these creates a thread pool, if I am not mistaken. Would there be value in injecting a single instance here or must there be 3 unique instances?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/StringifyFields.java
```diff
@@ -0,0 +1,152 @@
...
+ * <p>
+ * Configuration: {@code fields} — comma-separated list of top-level value field names to stringify.
+ * <p>
+ * Error behavior: if a targeted field cannot be serialized to JSON, the SMT logs the offending
+ * record's context (field name, schema type/name, value class and value) at ERROR level and
+ * re-throws a {@link ConnectException} (fail-fast). The stream halts and the offset is not
+ * committed past the record, so nothing is dropped silently; set {@code errors.tolerance=all} on
+ * the engine if you prefer to skip such records instead.
+ */
+public class StringifyFields<R extends ConnectRecord<R>> implements Transformation<R> {
```
Naros: I wonder if there could be value in placing this in `debezium-connect-plugins`?

Naros (commented):
I've left two final points below, if you could take a look.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/DefaultTokenHttpClient.java
wagnercsantos: Good catch — you're right that each `HttpClient` brings its own thread pool, and there's no reason for the REST path to have more than one. On the REST route there were two: the consumer's client for the record POSTs, plus one created inside `DefaultTokenHttpClient` for the token exchange.

Fixed by making `DefaultTokenHttpClient` accept an existing `HttpClient`, so `ZerobusRestChangeConsumer` now builds a single client and shares it with the token provider — one client, one pool, serving both the POSTs and the token exchange. The no-arg constructor is still there for the Kafka-route login handler, which has no client of its own to share.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/StringifyFields.java
wagnercsantos: Agreed — there's nothing Zerobus-specific about the mechanism, only the use case (a Delta `VARIANT` column), so it belongs in the shared transformations. Moved it to `io.debezium.transforms.StringifyFields` in `debezium-connect-plugins`, registered in that module's `Transformation` service file, and dropped it (plus its test and service registration) from the sink module. The javadoc is now phrased generically, mentioning `VARIANT` and JSON/JSONB columns as examples rather than framing it around Zerobus.

That part lives in a separate PR against `debezium` since it touches another repo: debezium/debezium#7782. Worth noting the merge order — this PR removes the SMT from the sink, so it should land together with (or after) that one, otherwise there'd be a window where the transformation doesn't exist anywhere. The sink itself doesn't reference the class (it's user-configured via `debezium.transforms.*`), and `debezium-connect-plugins` is already a transitive compile dependency of `debezium-server-core`, so it stays on the classpath either way.

wagnercsantos:
Pushed two commits that finish the observability story for this sink.

**Liveness and connection metrics.** `MilliSecondsBehindSource` freezes at its last value
when the pipeline stops, so on its own it cannot tell an idle source from a stalled sink.
Added `MilliSecondsSinceLastEvent`, which keeps growing while the sink forwards nothing,
and `Connected` — both mirroring the attributes of the same name on the source connectors.
The counters now also have a concurrency test: the class is annotated `@ThreadSafe` and is
updated from more than one thread, but nothing asserted that before.

**Metrics on the Kafka route.** That route writes through the generic `kafka` sink, so this
sink is never invoked and the metrics the other two routes expose were simply absent there.
A `ProducerInterceptor` is the extension point the Kafka producer already provides, so it
needs no change to the `kafka` sink; registering it reports the same attributes under
`server=kafka`. Two attributes read differently there, and the javadoc says so:
`TotalFlushes` counts broker acknowledgements, which is that route's durability boundary,
and `TotalRecordsSkipped` stays `0` because the generic sink forwards everything.

Since `onSend` runs per record, the extraction is kept off the allocation path: the
operation and source timestamp come from the record headers when `add.headers` is
configured, otherwise from a single-pass scan of the serialized value. Latency is sampled
rather than timed on every record — `System.nanoTime()` turns out to be an order of
magnitude more expensive when the producer's network thread reads the clock concurrently,
and that cost dominated everything else the interceptor did. The first record is always
timed so a low-volume pipeline still reports a duration, and all counters stay exact.

Tested end to end against a real workspace on the Kafka route, matching the counters and
the per-operation split against the Delta table, plus a 20k-event run to see the metrics
under load. There is also a gated integration test alongside the existing gRPC one, which
skips unless the credentials are in the environment.

One thing found along the way is unrelated to this sink and went to its own issue,
debezium/dbz#2380: the Prometheus exporter config that Debezium Server ships has no rule
matching the object name sinks use, so their metrics are exported with the task id inside
the `name` label. It affects the JDBC sink the same way. Fix in
debezium/debezium-server#301.

omkarmehta06:
Hi @wagnercsantos and @Naros — I reviewed the current PR head in detail against the ZeroBus sink we have been validating for NVIDIA. I think this PR is the right upstream foundation for our use case: it already provides the native Debezium Server integration, the GA ZeroBus SDK route, REST and Kafka-compatible options, typed JSON normalization, OAuth, SDK recovery controls, flush-before-commit ordering, JMX metrics, and broad tests. Building on this PR would let us contribute the NVIDIA requirements upstream instead of maintaining a competing sink.

The important clarification is that this is already a native Debezium Server sink. Our gap is not the runtime; it is the additional raw CDC data and operational contract we need.

| Area | PR #294 today | NVIDIA requirement / contribution |
|---|---|---|
| Debezium Server integration | Native `@Named("zerobus")` gRPC sink, plus native REST and Kafka-compatible routes | Keep this module and consumer as the single integration point; no competing module or second `zerobus` consumer |
| Data contract | Schema-driven typed Delta rows through `ZerobusTypeSystem`, normally with `ExtractNewRecordState` | Add a mutually exclusive raw-envelope mode that preserves the CDC event and leaves typed materialization downstream |
| Formats | Typed-row JSON | Add raw-envelope JSON and Protobuf in this contribution. Typed-row Protobuf can be a subsequent PR |
| Event fidelity | Operation and source timestamp are extracted for mapping/metrics; the typed path intentionally reshapes the event | Preserve destination, partition, key, value, headers, operation, source partition/offset, and an idempotency key in the stored envelope |
| Deletes and tombstones | Tombstones are skipped; deletes are represented through the configured unwrap/rewrite behavior | Distinguish delete events from tombstones and support explicit `event` or `drop` tombstone handling |
| Durability boundary | Records are enqueued individually, touched streams are flushed, and Debezium records are committed after all flushes | Retain flush-before-commit and add batched `ingestRecordsOffset(...)` plus `waitForOffset(...)`; commit only records covered by the durable acknowledgement |
| Metrics timing | `TotalRecordsIngested` is incremented when `ingestRecordOffset(...)` returns, before the later stream flush | Record acknowledged success after the durability barrier so a failed flush cannot over-count durable records; add acknowledged bytes/offset, retry, ambiguous-delivery, and oversize counters |
| Retry ownership | SDK recovery is configurable and sink failures propagate without an internal full-batch replay | Classify confirmed pre-delivery failures separately from outcomes where delivery may have started; never add an outer resend for an ambiguous outcome |
| Deduplication | Documentation recommends downstream deduplication using source LSN/offset carried by the selected payload shape | Emit a deterministic idempotency key derived from the Debezium source position while retaining the complete source-position map |
| Stream resources and ordering | One JSON stream per table is retained in an unbounded `HashMap`; touched streams are flushed independently | Add a configurable LRU stream bound and preserve source submission order when batches contain interleaved target tables |
| Record limits | No encoded-record size guard before SDK ingestion | Enforce a configurable encoded byte limit for JSON and Protobuf before delivery, with source-position context in the failure |
| Filtering | Filtering/routing can be handled with Debezium transforms and the existing stream-name mapper | Keep the existing mapper; add optional sink-native include filters needed when the raw landing contract must retain the original event |
| Standalone packaging | The sink JAR is included, but `zerobus-ingest-sdk` is optional and was absent from the `sink-databricks-zerobus` custom distribution I built; the gRPC route therefore needs the SDK added manually | Make the selected gRPC custom distribution self-contained with the ZeroBus SDK and Protobuf runtime, and retain the JNI packaging check |
| Verification | Strong typed mapping, OAuth, recovery, REST, Kafka, metrics, and gRPC tests | Add raw-envelope JSON/Protobuf, connector-offset compatibility, idempotency, tombstone, failure-ordering, size-limit, LRU, and packaging tests |

The existing flush-before-commit design and typed JSON path should remain intact. I propose adding a mode boundary along these lines:

```properties
debezium.sink.zerobus.payload.mode=typed|envelope
debezium.sink.zerobus.record.format=json|protobuf
```

For this PR, `typed + json` remains the current behavior, while `envelope + json|protobuf` provides NVIDIA's lossless landing path. `typed + protobuf` can be handled in a subsequent PR without moving raw-envelope Protobuf out of this collaboration.

Would you be open to us contributing these changes directly to your PR branch? I can base signed commits on the current head and open a focused PR into `wagnercsantos:dbz-2279-zerobus-sink`, so both implementations and authorship remain visible in #294.

wagnercsantos:
Heads-up on a regression I introduced in `d8673b1a` and have just fixed in `8476d942`: `Verify debezium-server distribution (redis)` went red, and the cause was mine rather than flaky.

Adding `@Inject @CustomConsumerBuilder Instance<ZerobusSdk>` puts `ZerobusSdk` in the bean's signature, so Quarkus emits a `Class.forName("com.databricks.zerobus.ZerobusSdk")` in the generated `_Bean` constructor. That runs in `Arc.initialize()`, before any sink is selected — and the SDK was declared `<optional>`, so it was absent from a distribution that nonetheless contained this module. Result: `ClassNotFoundException` and the server failing to start **with any sink**, redis included.

The root cause was the `<optional>` flag: this was the only module in the repository using it, and it left the module in the distribution without its client. The fix drops it, so the SDK is a regular dependency and the distribution controls inclusion through `sink.databricks-zerobus.scope` like every other sink does. The profile no longer needs to request the SDK explicitly, so that block goes away too — the change is net negative in lines.

Verified locally with the same two checks the job performs: the default `-Passembly` distribution starts with `debezium.sink.type=redis` in 1.08s and `/q/health` returns 200, and a distribution built with `-Psink-databricks-zerobus` still ingests over gRPC against a real workspace (`ingested=9 (r=9) errors=0`). 129 unit tests green.

One consequence worth flagging: the SDK bundles per-platform JNI libraries, so the default distribution grows by ~39 MB (498 MB total). That is consistent with the sink jar already shipping there, and `pulsar-client` is precedent for a native-bearing client at `compile` scope — but if you would rather keep the default distribution lean, giving this sink a `test` default scope also works and I have it validated.

omkarmehta06:
Hi @wagnercsantos and @Naros, I pushed NVIDIA's agreed envelope contribution as `a686d03a`, based directly on `8476d942`.

It preserves the existing typed JSON path and adds JSON/Protobuf envelopes, deterministic source identity, explicit tombstones, encoded-size validation, sink-native filters, source-order-preserving table groups, and acknowledgement-gated offset commits.

Local verification passed the focused envelope suite (`9/9`) and the full Zerobus module suite (`140/140`). The DCO and commit-message checks are green. Cross Maven CI is currently waiting for maintainer approval; could a maintainer approve the workflow and review the new commit?

One follow-up before merge: the PR description and documentation PR #7725 still reflect the pre-envelope implementation and should be refreshed with the new configuration and behavior.

wagnercsantos:
@omkarmehta06 — reviewed `a686d03a` and then ran the whole envelope contract end to end against a real workspace. Because I forgot an orphan dependency. Everything holds up well: it slotted into the `ZerobusStreamHandle` seam without reworking `handleTyped()`, `ZerobusTypeSystem` stays off the envelope path, formatter and checkstyle are clean. Four things below, one of which needed a commit from me.

**Protobuf dependency — moved to the BOM in** `874aa3b8`**.** The dependency itself was the right call: the SDK exposes `DescriptorProto`, `Message` and `Parser` on its public API while declaring none of them at compile scope, so `createProtoStream` genuinely does not compile without it. That's a defect in the SDK's own POM and you were the first to hit it.

The distribution already resolves `protobuf-java` at 4.33.2, transitively through `debezium-server-qdrant`, and no module here declares a version inline; they all inherit from a BOM. I know that convention the hard way: I made the same mistake with `version.zerobus.sdk` in the original sink commit, and that orphaned property was still sitting in the module POM until this commit cleaned it up alongside yours. One decision — versions live in the BOM — applied to both our declarations.

For context, the connectors stay on Protobuf 3.25.5: `debezium-connector-postgres` and `debezium-connector-oracle` generate code from `.proto` files with `protoc` pinned to that same property, and two of those files are proto2. Moving them is a regeneration exercise, not a version bump. `debezium-server` and the connectors are separate artifacts, so 4.x here and 3.x there coexist; I'll flag it to @Naros so it reads as deliberate.

`tombstone.handling.mode=event` **was unreachable — implemented in** `4c5910f7`**.** This one I could not leave, because the option is documented and could never fire. `TombstoneSupportProducer` in `debezium-server-core` withholds tombstones unless a consumer reports the capability:

```java
return changeConsumerHolder.tombstoneSupport()
    .map(isSupported -> (CapturingTombstoneEvents) () -> isSupported)
    .orElseGet(() -> () -> false);        // default: no tombstones
```

The sink did not implement `tombstoneSupport()`, so an end-to-end run with `tombstones.on.delete=true` and `tombstone.handling.mode=event` landed the delete and no tombstone at all. Only four sinks in the repo implement it (two PubSub, two RabbitMQ), which is presumably why it isn't obvious.

Your mapper already handles the case correctly — `operation()` returns `TOMBSTONE` on a null value, and the `drop` branch works — so the missing piece was only asking the engine for them. It now reports the capability when both `envelope` and `event` are configured, leaving the typed and REST paths receiving none. One implementation detail worth flagging: I read the configuration directly there rather than from the field `@PostConstruct` populates, because the engine queries the capability while deciding what to capture and that is not ordered against bean initialisation — the same trap that gave me a start-up NPE with the `Instance<ZerobusSdk>` injection point earlier this week. After the change the tombstone lands with `operation=tombstone`, its key preserved and a null value.

**One request on** `filter.operations`**.** It takes `create`, `read`, `update`, `delete`, and the connector-side `skipped.operations` takes `c`, `u`, `d`, `t` — both live in the same properties file. I typed `filter.operations=c,u` out of habit and the sink refused at start-up, which is good behaviour but points at the mismatch: `Envelope.Operation` in `debezium-connector-common` defines these as single-character codes, and `skipped.operations` has used them for years. Would you consider accepting the short codes as the canonical form, keeping the long names as aliases? It's your API surface and it touches the docs, so I'd rather you made the call than change it myself.

**And one observation from the run.** I went looking at the landed rows at my delta table and `idempotency_key` came out bigger than I expected: 573 to 609 characters, averaging 588, about 18% of the size of the `value` column beside it. `encode()` Base64s each component, so the whole serialized key JSON ends up inside the identity string, and a composite primary key would grow it further. It is genuinely deterministic — 8 events, 8 distinct keys — so this is not a correctness point.

I did benchmark hashing before raising it, since it isn't free — the digest is an extra pass on top of building the canonical string, not a replacement. But the CPU cost is negligible next to what it saves on the wire and in the table, and this route is bound by round trips rather than local processing, so the trade looks worth making.

What a hash costs is readability: the key can currently be read to see the LSN and the operation. Though `source_position` is in the envelope regardless, so the debugging path arguably lives there rather than in the key. Was the readable form a deliberate choice?

**What the runs covered,** all against a live workspace with Protobuf 4.33.2 in the distribution:


|                                                  |                                                                                                                                                                                                                            |
| ------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `envelope` + `json`                              | `create`, `update`, `delete` and 5 `read` matching the sink metrics exactly; `value` carrying the full schema and payload; `source_position` carrying LSN, txId and messageType; the delete arriving with its `key` intact |
| `envelope` + `protobuf`                          | same counts, 8 distinct keys, the service decoding the bytes against your `DescriptorProto`                                                                                                                                |
| `max.record.bytes`                               | rejected a 5566-byte envelope at a 800-byte limit, reported the source position, stopped the connector, and left **zero rows** in the table — the "offsets are not committed" claim verified                               |
| `filter.operations` + `filter.destination.regex` | `ingested=2 (c=1 u=1 d=0)`, with 6 skips attributed to `operation` and 1 to `destination`                                                                                                                                  |
| `tombstone.handling.mode=event`                  | fails before `4c5910f7`, lands correctly after                                                                                                                                                                             |


One thing that cost me a run and is now in the docs: applying `ExtractNewRecordState` in envelope mode makes every row `operation=change`, because the SMT strips the `op` field and the mapper falls back. Worth stating explicitly since the typed mode wants that transformation and this one must not have it.

Two design questions still open, both genuine:

1. In `handleEnvelope()`, a batch with interleaved tables writes one group per contiguous run. If a later group fails, the earlier ones are already durable and no offset is committed, so the batch replays and those groups are re-sent. With at-least-once plus the idempotency key that's defensible — I just want to confirm it's the intended trade-off.
2. `configureEnvelopePath()` is called from both `connect()` and the top of `handleEnvelope()`. The `envelopeMapper != null` guard makes the second a no-op, but it's a per-batch check on the hot path. Is there a lifecycle case where `connect()` hasn't run first that I'm missing?

On docs: I've left the envelope options for you on [debezium#7725](https://github.com/debezium/debezium/pull/7725), since you know the reasoning behind each one — `max.record.bytes`, `json.flexible.fields.encoding`, `idempotency.mode`, `tombstone.handling.mode` and the seven `filter.*` options aren't in the options table yet, and the row contract of the envelope probably deserves its own subsection. 

What I did change there was two sentences of my own that your commit made false — they claimed the sink implemented `typed`/`json` only — plus a note that `ExtractNewRecordState` must not be applied in envelope mode, since that one cost me a run. =D

omkarmehta06:
@wagnercsantos, thank you for the detailed review and for validating the envelope paths against a live workspace. I reviewed `874aa3b8` and `4c5910f7`; both look good.

I pushed `9eeeafd8` to address the open points:

- `filter.operations` now accepts the canonical Debezium codes `c`, `r`, `u`, `d`, `t`, and `m`, while retaining long aliases. `t` remains truncate; tombstones use the explicit `tombstone` name. Truncate and message events are now preserved instead of falling back to `change`.
- `idempotency_key` is now `sha256:<64 hex characters>` over the canonical source identity. The readable connector position remains in `source_position`.
- The interleaved-table behavior is the intended at-least-once trade-off: if a later table group fails, an earlier acknowledged group may replay because the batch offsets remain uncommitted. A regression test now proves that boundary; downstream deduplication uses `idempotency_key`.
- I removed the redundant envelope configuration check from the per-batch path. `connect()` initializes it, and the unit harness now invokes the same initialization explicitly.
- The tombstone capability tests now invoke the consumer's actual `tombstoneSupport()` method.

Verification passed: `146/146` module tests, formatter validation, import sorting, and the Debezium OSS contribution checker.

For documentation PR #7725, my current contribution scope covers `debezium/debezium-server`. I can prepare the exact envelope option and row-contract text for you to apply there, or follow the separate approval path before editing that repository.

@Naros, once CI completes on this commit, this should be ready for maintainer review.

wagnercsantos:
@omkarmehta06 — re-ran the envelope contract end to end against `9eeeafd8` in a fresh schema, and all four fixes hold up.

`filter.operations=c,u,d,r` starts cleanly, which is the exact configuration that refused to start yesterday, and adding `TRUNCATE` and `MESSAGE` to the enum rather than letting them fall back to `change` is better than what I asked for. `idempotency_key` is now 71 fixed characters against 573–609 before, still one distinct key per event, with the readable position preserved in `source_position`. Both formats landed identically — `create`, `update`, `delete`, 2 `read`, zero errors, JSON and Protobuf matching each other row for row. Making the tombstone tests call the real `tombstoneSupport()` instead of mirroring the condition is also an improvement on what I wrote.

One observation, not a problem: with `tombstone` absent from the operations allowlist the tombstone is skipped, so the filter takes precedence over `tombstone.handling.mode`. That's the sensible ordering, and might be worth a sentence in the docs so nobody configures both and wonders where the tombstones went.

On the documentation: either path works for me, whichever is easier on your side. Send me the envelope option text and the row-contract wording and I'll apply it to #7725 with attribution to you in the commit message. Or, if you can get approval for the `debezium` repository, `wagnercsantos/debezium` is the fork behind that PR and I can grant you access there exactly as on the server one — a push to `dbz-2279-zerobus-docs` lands straight in #7725, same as here. Your call; I'd rather the words be yours either way, since you know the reasoning behind each option.

From my side this is done: 146 tests green locally, formatter, imports and checkstyle clean, and the envelope validated live in both formats. @Naros, ready for review whenever you have the time.

omkarmehta06:
@wagnercsantos, thank you. I would like to take the direct-contribution path so the documentation commit preserves my authorship and `Signed-off-by` trail. Please grant `omkarmehta06` push access to `wagnercsantos/debezium` for the `dbz-2279-zerobus-docs` branch.

Once access is active, I will start from the branch current head and add one focused signed commit covering the missing envelope options, row contract, operation/filter semantics, and tombstone/filter precedence. I will preserve the existing commit history without rewriting or squashing it.

wagnercsantos:
@omkarmehta06 — done, invite sent for `wagnercsantos/debezium` 
Current head is `9a52f37116` and #7725 is `mergeable: clean`, so a fast-forward push lands straight in the PR.

wagnercsantos:
@kmos should I solve the conflict?

kmos:
> @kmos should I solve the conflict?

Yes, thanks!

wagnercsantos:
@kmos done — merged `upstream/main` into the branch in `ead0a38a`. The only conflict was in `debezium-server-dist/pom.xml`, where dbz#2433 added the storage/scripting scope properties right where this PR adds the `sink.databricks-zerobus.scope` one; I kept both. Rebuilt the distribution with the sink profile to confirm it still assembles and the CDI bean generates. It's `CLEAN` now.

Naros:
Hi @wagnercsantos we can't accept PRs that have merge commits. Could you remove the merge commit and instead do a rebase against `main`?

wagnercsantos:
@Naros / @kmos done — rebased against `main` in `b4c8d423`, the two merge commits are gone and the history is now linear (15 commits).

While resolving the rebase I picked up the `debezium-server-dist/pom.xml` changes from dbz#2433 (the storage/scripting scope properties) and dbz#2318 (Db2 in the assembly), so the sink's `sink.databricks-zerobus.scope` property sits alongside them without duplication. Rebuilt the distribution with the sink profile to confirm it still assembles and the CDI bean generates.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumer.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private void maybeLogMetrics() {
+        final int interval = config.getMetricsLogInterval();
+        if (interval > 0 && ++batchesSinceLog >= interval) {
+            batchesSinceLog = 0;
+            metrics.logMetricsSummary();
+        }
+    }
+
+    private void post(String table, String json) throws IOException, InterruptedException {
```
kmos: Could you please extract this method in another class like `DefaultTokenHttpClient`?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumer.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Debezium Server sink that writes change events into Databricks Zerobus Ingest over the REST API
+ * ({@code POST <uri>/zerobus/v1/tables/<catalog.schema.table>/insert}). Events are serialized as
+ * flat JSON by Debezium and posted one per request with an OAuth bearer token.
+ * <p>
+ * This route needs neither the Kafka ingress flag nor a persistent connection, so it suits
+ * serverless / edge deployments. Like the other routes it is at-least-once; deduplicate downstream.
+ */
+@Named("zerobusrest")
```
kmos: it could be better to use naming `zerobus-rest`? @Naros @wagnercsantos WDYT?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusEventFilter.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.debezium.DebeziumException;
+import io.debezium.engine.Header;
+import io.debezium.runtime.BatchEvent;
+
+/** Applies optional sink-native filters before an envelope is routed or serialized. */
+final class ZerobusEventFilter {
```
kmos, line 20: If I understand correctly, you want to filter events. Did you try Debezium Filter SMT ? https://debezium.io/documentation/reference/stable/transformations/filtering.html

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
```diff
@@ -0,0 +1,678 @@
...
+            metrics.recordError();
+            throw e;
+        }
+        catch (Exception e) {
+            metrics.recordError();
+            throw new DebeziumException("Failed to durably ingest Zerobus envelope group for table '" + table + "'", e);
+        }
+    }
+
+    private void writeJsonEnvelopeGroup(String table, List<ZerobusEnvelope> records) throws Exception {
```
kmos: Could you please extract all the serialization logic into a different class using a strategy pattern and initialize only one based on cofiguration (protobuf/Json)

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumer.java
Naros: I agree, and we already have the precedent in the NATS consumer with `nats-streaming` and `nats-jetstream`.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
```diff
@@ -0,0 +1,678 @@
...
+        for (String table : touchedTables) {
+            try {
+                streams.get(table).flush();
+            }
+            catch (Exception e) {
+                metrics.recordError();
+                throw new DebeziumException("Failed to flush Zerobus stream for table '" + table + "'", e);
+            }
+        }
+        metrics.flushed((System.nanoTime() - flushStartNanos) / 1_000_000L);
```
Naros: Should this be guarded by `!touchedTables.isEmpty()` to avoid setting the flushed metrics to 0 when there were no tables touched?

Naros (commented):
Hi @wagnercsantos I found just one more item, not sure you think it's important.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusEventFilter.java
wagnercsantos: @kmos the Filter SMT runs in the transformation chain, ahead of the sink, and needs a scripting engine (Groovy/GraalJS). This filter runs *inside* the sink's envelope path, where the record is still the full change-event envelope (op, source, headers) — the same reason `ExtractNewRecordState` can't be applied in envelope mode; an SMT there would alter or strip the envelope the sink is meant to preserve. It's a lightweight regex over the already-parsed event, with no extra dependency. @omkarmehta06 authored this path and can add anything I've missed.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
wagnercsantos: @kmos good suggestion. There are already two serializer classes (`ZerobusJsonEnvelopeSerializer` / `ZerobusProtobufEnvelopeSerializer`); the current dispatch is a single `if` on `record.format`. A full strategy pattern would need to be generic over the payload type — JSON produces `String`, Protobuf `byte[]`, and that split runs down to two separately-typed stream maps — so it's a mid-size refactor of the core write path, and the gain would be mostly stylistic since there's no functional change. Would it work to take it as a follow-up PR so this one stays focused? @omkarmehta06 owns this path — deferring to you on whether you'd prefer to fold it in here or handle it separately.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumer.java
wagnercsantos: @kmos done in `1dfa2322` — extracted the record-insert HTTP concern into a new `ZerobusRestClient`, the counterpart to the existing `DefaultTokenHttpClient` (token exchange). The consumer now holds the batch loop and delegates the request wiring; both share a single `HttpClient`.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusRestChangeConsumer.java
wagnercsantos: @kmos @Naros renamed to `zerobus-rest` in `1dfa2322`, following the `nats-streaming` / `nats-jetstream` precedent. Config prefix is now `debezium.sink.zerobus-rest.*` and `debezium.sink.type=zerobus-rest`; the docs are updated in #7725 to match, and I validated both routes end-to-end.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
wagnercsantos: @Naros done in `f7013861` — guarded the flush metric with `!touchedTables.isEmpty()`, so a batch with no touched tables (all tombstones/skipped) no longer records a 0 ms flush.

Naros (approved):
I think this LGTM from my PoV, just waiting on CI.

wagnercsantos:
@Naros, if you agree, I can open a new issue to work with the serialization logic as we mentioned before

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusEventFilter.java
kmos: Interesting. Is there any possibility of making this filter reusable across other sinks? From what I can see in the code, there are some sink-specific rules. Could these be generalized or made more consistent with the Debezium specification?

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusChangeConsumer.java
wagnercsantos: @kmos done in `04449078`. Extracted the serialization into a `ZerobusEnvelopeSerializer<P>` strategy: it serializes the envelope to its payload type, reports the encoded size, and opens the stream that accepts that encoding — the three things that varied by format, bound together by `P`. A factory instantiates only the one the configured `record.format` selects, so we no longer build both.

With that, the consumer holds a single `envelopeSerializer` and a single stream map instead of one field and one map per format, and the batch write path (`ingestEnvelopeGroup<P>`) is generic — serialization, size check, ingest, eviction and close no longer branch on JSON vs Protobuf. The one format decision left is the single `create()` at startup, which is unavoidable since the payload types (`String` vs `byte[]`) diverge into differently-typed streams.

No functional change. I re-validated end-to-end against a live workspace on both `record.format=json` and `record.format=protobuf` — snapshot read plus live create/update/delete/tombstone landing correctly in Delta.

@omkarmehta06 — heads-up, this touches the envelope serialization path you wrote. It's purely structural (nothing changes at runtime), and I mainly did it to get this thread cleared so the PR can move to review sooner rather than sit waiting. Totally fine to tweak it or take it over if you'd rather — just didn't want to leave your code changed without flagging it to you.

debezium-server-databricks-zerobus/src/main/java/io/debezium/server/databricks/zerobus/ZerobusEventFilter.java
wagnercsantos: @kmos yeah — mechanically most of this is generic. The destination / operation / header / value-pointer regex rules don't know anything about Zerobus; the only sink-specific bits are that it reads `ZerobusChangeConsumerConfig` and asks the envelope mapper for the operation. So a shared sink-side filter is definitely feasible: lift the rule engine into `debezium-server-core` (or a shared helper), drive it off generic `debezium.sink.filter.*` keys, and let each consumer hand it the destination/op/headers/value.

The catch is that it's a cross-cutting change — it lands in server-core and effectively defines a small filtering contract for every Debezium Server sink, plus it overlaps with the Filter SMT you mentioned earlier. That feels like a maintainer's call on whether it belongs in the framework, rather than something to introduce through this sink's PR — I'd rather not stretch #294 into an SPI change on my own.

Proposal: keep the filter self-contained here (it's tested and localized) and take "generalize into a shared server-side filter" as a follow-up. @omkarmehta06 wrote this path, so he should weigh in first. Omkar — if you're swamped, just throw it my way and I'll pick it up in a separate PR, no worries; otherwise it's all yours. And @Naros — would a shared sink-side filter in server-core be something you'd want, or would you rather sinks lean on the Filter SMT for this?

kmos: Yeah @wagnercsantos the idea was to use the [Filter APIs for the quarkus extension](https://debezium.io/documentation/reference/3.6/integrations/quarkus-debezium-engine-extension.html#_by_custom_filter_strategy) as baseline. You can refactor the filter logic in a separate PR once we've adapted the Quarkus filter to Debezium Server.

kmos (approved):
@wagnercsantos @omkarmehta06 Nice work! If you have a production use case related to this sink, we would be delighted to hear about it and feature it in our [Debezium Showcase](https://debezium.io/blog/2026/06/24/debezium-monthly-community-showcase/). We can arrange it using [sessionize](https://sessionize.com/debezium-community-showcase).

Naros:
Hi @wagnercsantos could we get a quick rebase, it seems there's a conflict with the BOM.

wagnercsantos:
> @wagnercsantos @omkarmehta06 Nice work! If you have a production use case related to this sink, we would be delighted to hear about it and feature it in our [Debezium Showcase](https://debezium.io/blog/2026/06/24/debezium-monthly-community-showcase/). We can arrange it using [sessionize](https://sessionize.com/debezium-community-showcase).

Hey @kmos, I have a customer waiting for the PR to be merged so they can deploy this. We are deploying a POC environment.

Naros:
Hi @wagnercsantos sorry for the late merge, Giovanni and I had a long morning. 

Thanks for all the work on this you and @omkarmehta06 did, great addition to the portfolio. :rocket:

wagnercsantos:
@Naros no apology needed at all — thank *you* and Giovanni for the thorough reviews and for landing it. 🎉 It's been a genuinely great collaboration with @omkarmehta06 and the whole Debezium team, and I'm thrilled to see the Zerobus sink in the portfolio.

One thing I want to clear up: my note to @kmos about a customer waiting wasn't a nudge to hurry the merge — sorry if it read that way! It was really an answer to the Showcase invite. There *is* a real use case behind it (we're standing up a POC deployment), and I'll check with the customer and suggest they take part — it'd make a great success story for the Community Showcase.

Thanks again to you and Giovanni — 🚀

kmos:
> @Naros no apology needed at all — thank _you_ and Giovanni for the thorough reviews and for landing it. 🎉 It's been a genuinely great collaboration with @omkarmehta06 and the whole Debezium team, and I'm thrilled to see the Zerobus sink in the portfolio.
> 
> One thing I want to clear up: my note to @kmos about a customer waiting wasn't a nudge to hurry the merge — sorry if it read that way! It was really an answer to the Showcase invite. There _is_ a real use case behind it (we're standing up a POC deployment), and I'll check with the customer and suggest they take part — it'd make a great success story for the Community Showcase.
> 
> Thanks again to you and Giovanni — 🚀

@wagnercsantos good luck for the POC! Feel free to reach us in [#zulip](https://debezium.zulipchat.com/#narrow/channel/302533-dev)

wagnercsantos:
@kmos following up on making the sink filter reusable across sinks, using the Quarkus extension's filter API as the baseline you suggested.

Good news on the groundwork: Debezium Server already depends on the full `debezium-quarkus-engine` extension (not just the SPI), and the envelope dispatch choke point — `DefaultChangeConsumer.handleBatch(CapturingEvents<BatchEvent>)` — is already `@Capturing`-annotated. So the `CapturingFilterStrategy` machinery (invoker generation + `CapturingFilterInvokerRegistry`) is already on the Server classpath and running; today `handleBatch` just uses the default accept-all strategy.

That means "adapting the Quarkus filter to Debezium Server" looks less like building a filter framework and more like: provide a **config-driven `CapturingFilterStrategy`** in `debezium-server-core` (reading generic `debezium.sink.filter.*` keys — destination/operation/header/value regex + malformed mode, the same rules the Zerobus filter has today) and attach it via `@Capturing(filter = ...)` at the dispatch point. Then a **second PR** drops `ZerobusEventFilter` and moves the sink onto the shared keys.

One semantic wrinkle worth your view: `shouldCapture` receives `CapturingEvent<SourceRecord, SourceRecord>` — the raw Connect `Struct` — whereas the current `filter.value.json.pointer` rule matches against the *serialized* JSON value. So the value-pointer rule either serializes first (keeps today's semantics) or moves to operate on the `Struct` (more canonical). The destination/operation/header rules map over cleanly.

I'll take this on end to end: PR-1 in `debezium-server-core` (the config-driven `CapturingFilterStrategy` wired at the dispatch point), then PR-2 refactoring `ZerobusEventFilter` onto the shared `debezium.sink.filter.*` keys — with @omkarmehta06 looped in as reviewer on PR-2, since he authored that path. Since PR-1 is cross-cutting in core, I'll align the design with you before coding.

I've opened debezium/dbz#2614 to track this.

omkarmehta06:
Thanks, @wagnercsantos, for taking this forward and keeping me involved. The two-PR approach works for me, pending @kmos’s agreement on the core design.
For the Zerobus migration, I’d prefer to preserve the existing serialized-JSON semantics for value-pointer rules and malformed-value handling. Please also retain compatibility with the existing filter configuration, or provide a clear migration path, so users can adopt the shared filter without unexpectedly changing which records are accepted or dropped.
Happy to help with either PR if there’s a piece I can pick up. Reviewing the Zerobus migration works for me too. Let me know what would be most helpful!

PR #299, opened 2026-08-03
debezium/dbz#2343 Add missing dependencies

vjuranek:
Fixes https://github.com/debezium/dbz/issues/2343


vjuranek:
This is actually a bad fix to add dependecies on the arifact using it. We should just remove `provided` scope from these artifacts in `debezium-storage-redis` module in main DBZ repo.

PR #300, opened 2026-08-03
debezium/dbz#2343 Add missing storage dependencies

vjuranek:
Fixes https://github.com/debezium/dbz/issues/2343

This seems to be the best option - `debezium-storage-api` and `debezium-storage-common` are used by several other storage modules, so it's not good to bundle them in these modules multiple time. Add them as a dependency into `debezium-server-code` so that they are available to all other server modules without need to declare them in all the modules.


Naros (approved):

Naros:
Applied, thanks @vjuranek :tada:

PR #301, opened 2026-08-07
debezium/dbz#2380 Export sink connector metrics with correct Prometheus labels

### LINKED ISSUE #2380: Prometheus JMX Exporter mislabels sink connector metrics in Debezium Server
## What

The `metrics.yml` file that Debezium Server ships for the bundled Prometheus JMX Exporter has no rule that matches the object name sink connectors use, so their metrics are exported with a malformed `name` label and no `task` label.

Sink connectors register their MXBeans as:

```
debezium.<plugin>:type=connector-metrics,context=sink,server=<name>,task=<id>
```

The rules in `metrics.yml` cover two orderings of those keys:

```yaml
# has task, but expects server BEFORE context
- pattern: "debezium.([^:]+)<type=connector-metrics, server=([^,]+), task=([^,]+), context=([^>]+)>([^:]+)"

# has context before server, but no task
- pattern: "debezium.([^:]+)<type=connector-metrics, context=([^,]+), server=([^>]+)>([^:]+)"
```

A sink object name has `context` before `server` **and** carries `task`, so only the second rule matches, and its `server=([^>]+)` group captures greedily up to `>`.

## Effect

The `task` key ends up inside the `name` label:

```
debezium_metrics_TotalNumberOfWrites{context="sink",name="jdbc-sink, task=0",plugin="jdbc"} 42.0
```

Compare a source connector in the same scrape, which is labelled correctly:

```
debezium_metrics_TotalNumberOfEventsSeen{context="streaming",name="inventory",plugin="postgres"} 42.0
```

Consequences for anyone scraping Debezium Server with Prometheus:

- the `name` label is not the connector name, so grouping and alerting by connector break;
- there is no `task` label, so per-task series cannot be distinguished;
- the label value contains `, ` — awkward to match in PromQL and inconsistent with every other Debezium series.

The metrics are correct over JMX itself (JConsole, jmxterm); only the Prometheus export is affected.

## Affected connectors

Any sink whose MXBean follows the shared convention. Confirmed by reading the code:

- `JdbcSinkConnectorMetrics` — `debezium.jdbc:type=connector-metrics,context=sink,server=%s,task=%s`
- the Databricks Zerobus sink proposed in debezium/debezium-server#294, which follows the same convention

## Proposed fix

Add one rule for the sink ordering, placed before the existing generic rule so that it wins:

```yaml
- pattern: "debezium.([^:]+)<type=connector-metrics, context=([^,]+), server=([^,]+), task=([^>]+)>([^:]+)"
  name: "debezium_metrics_$5"
  type: GAUGE
  labels:
    plugin: "$1"
    name: "$3"
    task: "$4"
    context: "$2"
```

After the change, the same scrape yields:

```
debezium_metrics_TotalNumberOfWrites{context="sink",name="jdbc-sink",plugin="jdbc",task="0"} 42.0
```

which matches the label shape the source connectors already produce.

## How this was verified

Ran Debezium Server with `JMX_EXPORTER_PORT` set and scraped `/metrics`, before and after the change, comparing a sink series against a source series in the same output. The malformed `name="…, task=0"` becomes `name="…"` plus a separate `task="0"`, and all counter values are unchanged.

The change above is the complete fix and is already verified locally. I will open a PR against `debezium-server` referencing this issue.

wagnercsantos:
Fixes debezium/dbz#2380.

Sink connectors register their MXBeans as
`debezium.<plugin>:type=connector-metrics,context=sink,server=<name>,task=<id>`, an
ordering that none of the existing rules in `metrics.yml` matches: one expects `server`
before `context`, and the other has no `task` and captures `server` greedily up to `>`.
Only the latter matches a sink, so the `task` key ends up inside the `name` label and no
`task` label is produced.

Before:

```
debezium_metrics_TotalNumberOfWrites{context="sink",name="jdbc-sink, task=0",plugin="jdbc"} 42.0
```

After:

```
debezium_metrics_TotalNumberOfWrites{context="sink",name="jdbc-sink",plugin="jdbc",task="0"} 42.0
```

which is the label shape the source connectors already produce.

The change adds a single rule for the sink ordering, placed before the generic one so
that it takes precedence. Existing rules are untouched, so source connector series are
unaffected.

Verified by running Debezium Server with `JMX_EXPORTER_PORT` set and scraping `/metrics`
before and after, comparing a sink series against a source series in the same output:
the malformed `name` is corrected, a `task` label appears, and all counter values are
unchanged.


wagnercsantos:
Hi @Naros — would you have a moment to review this one? It's a small fix (+8/−0 in metrics.yml): the Prometheus JMX exporter config wasn't capturing the context and task labels that the sink connector metrics expose, so those series landed without the labels needed to tell connectors apart. This adds the mapping rules for them.

The branch is rebased on current main with linear history, and it's CLEAN follow the same rule at #294. Happy to adjust anything.

vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @wagnercsantos !

wagnercsantos:
tks @vjuranek!!!!

PR #304, opened 2026-08-13
debezium/dbz#2416 Native build is broken for JDBC sink due to missed dialect

### LINKED ISSUE #2416: Native build is broken for JDBC sink due to missed dialect
missed dialects:
- StarRocksDatabaseDialect
- SingleStoreDatabaseDialect

kmos:
closes debezium/dbz#2416


debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/quarkus/JdbcSinkReflectionConfiguration.java
```diff
@@ -22,6 +25,9 @@
         SqlServerDatabaseDialect.SqlServerDatabaseDialectProvider.class,
         Db2DatabaseDialect.Db2DatabaseProvider.class,
         Db2iDatabaseDialect.Db2iDatabaseProvider.class,
+        StarRocksDialectResolver.class,
+        SingleStoreDatabaseDialect.class,
```
mfvitale, line 30: is `SingleStoreDatabaseDialectProvider` missing?

mfvitale:
@kmos Do you think is there a way (maybe leveraging on archunit) to add a rule that checks possible things to be added to the reflection configuration?

debezium-server-jdbc/src/main/java/io/debezium/server/jdbc/quarkus/JdbcSinkReflectionConfiguration.java
kmos: I have done a runtime execution and I didn't receive the error due to this missing class but I agree that it's weird.

kmos:
> @kmos Do you think is there a way (maybe leveraging on archunit) to add a rule that checks possible things to be added to the reflection configuration?

Nice idea, but I don't see how we could proactively load the dialects. These classes are registered through reflection, so their existence is not known at compile time. In any case, you would still need to maintain a list of the available dialects.

@rk3rn3r do you have any suggestion?

mfvitale:
> > @kmos Do you think is there a way (maybe leveraging on archunit) to add a rule that checks possible things to be added to the reflection configuration?
> 
> Nice idea, but I don't see how we could proactively load the dialects. These classes are registered through reflection, so their existence is not known at compile time. In any case, you would still need to maintain a list of the available dialects.
> 
> @rk3rn3r do you have any suggestion?

At least creating a rule that detect the addition of a new dialect so the Ci fails asking users to add it to that list?

kmos:
Also @Naros [mentioned](https://github.com/debezium/dbz/issues/2416#issuecomment-5280884506) the need to verify the build.

rk3rn3r:
> These classes are registered through reflection, so their existence is not known at compile time.

I don't know if this is true. At the moment we should exactly know which dialects exists. And maybe it is an acceptable option to generate the list for those (I've added a comment to the conversation in debezium/dbz#2416). Maybe it is acceptable to have this limitation, it is only for the native build, right? We could later add other ways for dynamically linking custom dialect implementations (which we might see not that often at all)?

kmos:
> > These classes are registered through reflection, so their existence is not known at compile time.
> 
> I don't know if this is true. At the moment we should exactly know which dialects exists. And maybe it is an acceptable option to generate the list for those (I've added a comment to the conversation in [debezium/dbz#2416](https://github.com/debezium/dbz/issues/2416)). Maybe it is acceptable to have this limitation, it is only for the native build, right? We could later add other ways for dynamically linking custom dialect implementations (which we might see not that often at all)?

I mean that when someone adds a new dialect to the package, it's only loaded at runtime (via reflection). In any case, this only affects native builds and the small number of dialects we currently have, so it's not a big trouble.

I've added an ArchUnit rule that fails if it detects a new class implementing a dialect in the JDBC package that is not already included in the list. @mfvitale let me know if it's what you suggested.

vjuranek (approved):
LGTM

vjuranek (commented):
You just need to sort the import @kmos

mfvitale:
> I've added an ArchUnit rule that fails if it detects a new class implementing a dialect in the JDBC package that is not already included in the list. @mfvitale let me know if it's what you suggested.

I tought it shoudl be added in https://github.com/debezium/debezium/blob/main/support/archunit/src/main/java/io/debezium/testing/archunit/DebeziumArchitectureTest.java

kmos:
> > I've added an ArchUnit rule that fails if it detects a new class implementing a dialect in the JDBC package that is not already included in the list. @mfvitale let me know if it's what you suggested.
> 
> I tought it shoudl be added in [debezium/debezium@`main`/support/archunit/src/main/java/io/debezium/testing/archunit/DebeziumArchitectureTest.java](https://github.com/debezium/debezium/blob/main/support/archunit/src/main/java/io/debezium/testing/archunit/DebeziumArchitectureTest.java?rgh-link-date=2026-08-14T10%3A58%3A51Z)

I don't know, it means that we should add the sink as dependency. If it's possible to invert the dependency: create a kind of `debezium-server-test-suite` and import in the sinks

mfvitale (approved):

mfvitale:
> > > I've added an ArchUnit rule that fails if it detects a new class implementing a dialect in the JDBC package that is not already included in the list. @mfvitale let me know if it's what you suggested.
> > 
> > 
> > I tought it shoudl be added in [debezium/debezium@`main`/support/archunit/src/main/java/io/debezium/testing/archunit/DebeziumArchitectureTest.java](https://github.com/debezium/debezium/blob/main/support/archunit/src/main/java/io/debezium/testing/archunit/DebeziumArchitectureTest.java?rgh-link-date=2026-08-14T10%3A58%3A51Z)
> 
> I don't know, it means that we should add the sink as dependency. If it's possible to invert the dependency: create a kind of `debezium-server-test-suite` and import in the sinks

> we should add the sink as dependency.

As we did with `debezium-connector-common` to have a single place where we define this kind of rules. In any case this is not a blocker, I'm also fine with the current solution. It's up to you.

PR #312, opened 2026-09-01
debezium/dbz#2167 Map TotalNumberOfReadEventsSeen in the JMX exporter configs

ragulshanmugam:
Follow-up to debezium/debezium#7896, which added the `TotalNumberOfReadEventsSeen` MBean attribute to `CommonEventMeter`. This maps that attribute in both JMX exporter configs, so snapshot event throughput becomes visible to Prometheus and OTel consumers.

This is point 3 of debezium/dbz#2167.

### What changed

`debezium-jmx-config.yaml` gains a `TotalNumberOfReadEventsSeen` mapping alongside the existing create, update and delete entries, emitting `event.count` with `debezium.event.type: const(read)`.

`debezium-jmx-config-legacy.yaml` gains the equivalent passthrough entry, matching the shape its neighbouring counters use.

Each file has three rule blocks. They differ only by MBean ObjectName shape (`database=*`, `task=*`, `server=*`) rather than by phase, so the mapping is added to all three. The snapshot and streaming distinction is already carried as a label through `debezium.context: param(context)`.

### Note on streaming

Read events are only emitted during a snapshot, so `debezium_event_type="read"` reports 0 under `debezium_context="streaming"`. That mirrors how create, update and delete already behave under `debezium_context="snapshot"`. All four attributes are inherited from `CommonEventMetricsMXBean` and are therefore exposed on both MBeans.

### Testing

I did not run tests for this change. It is a resource-only configuration mapping with no test surface in this repo. I confirmed both files still parse as YAML and that the added blocks match the structure of the surrounding entries.

### AI assistance

I used Claude to help trace the metrics dispatch chain in debezium core and to review this diff before opening it. The mapping itself follows the outline in debezium/dbz#2167, and I verified the attribute names and ObjectName patterns against the source myself.


vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @ragulshanmugam !

PR #313, opened 2026-09-03
debezium/dbz#1582 Add TLS/mTLS support for the rabbitmqstream sink

### LINKED ISSUE #1582: TLS/mTLS support for RabbitMQ Stream sink (rabbitmqstream)
## Bug report

**What Debezium connector do you use and what version?**

Debezium Server 3.4.0 with rabbitmqstream sink

---

**What is the connector configuration?**
```
    #   connection.host: rabbitmq-host
    #   connection.port: "5551"
    #   connection.username: debezium
    #   connection.password: ${rabbitmq_password}
    #   connection.ssl.enabled: "true"
    #   connection.ssl.rootCert: /debezium/external/rabbitmq-tls/ca.crt
    #   connection.ssl.certFile: /debezium/external/rabbitmq-tls/tls.crt
    #   connection.ssl.keyFile: /debezium/external/rabbitmq-tls/tls.key
```
---

**What is the captured database version and mode of deployment?**

PostgreSQL 17, deployed on Kubernetes. RabbitMQ 4.2.0 cluster with mTLS enabled on the Stream protocol (port 5551).

---

**What behavior do you expect?**

I expect to be able to configure TLS/mTLS for the rabbitmqstream sink, similar to how the rabbitmq (AMQP) sink supports TLS via ConnectionFactoryConfigurator with PEM

---

**What behavior do you see?**

There is no way to configure TLS/mTLS for the `rabbitmqstream` sink. The `connection.*` passthrough properties do not seem to support TLS configuration for the Stream client.

When connecting to a TLS-enabled RabbitMQ Stream port, the connection fails with:
```
com.rabbitmq.stream.impl.ConnectionStreamException: Connection is closed
```

RabbitMQ logs show:
```
{tls_alert,{unexpected_message,"TLS server: In state hello at tls_record.erl:532 generated SERVER ALERT: Fatal - Unexpected Message\n {unsupported_record_type,0}"}}
```

This indicates the client is sending plaintext to a TLS port.

---

**Do you see the same behaviour using the latest released Debezium version?**

Yes, tested with Debezium Server 3.4.0.Final.

---

**Do you have the connector logs, ideally from start till finish?**

_You might be asked later to provide [DEBUG/TRACE](https://debezium.io/documentation/reference/stable/operations/logging.html) level log._

```
{"level":"WARN","message":"The parameters connection.host and connection.port are deprecated, please use rabbitmqstream.host and rabbitmqstream.port moving forward."}
{"level":"INFO","message":"Using connection to rabbitmq.example.com:5551"}
{"level":"ERROR","message":"Failed to start application","exception":{"exceptionType":"io.debezium.DebeziumException","message":"com.rabbitmq.stream.impl.ConnectionStreamException: Connection is closed"}}
```
---

**How to reproduce the issue using our [tutorial](https://github.com/debezium/debezium-examples/tree/main/tutorial) deployment?**

Deploy RabbitMQ with TLS/mTLS enabled on the Stream protocol (port 5551)
Configure Debezium Server with rabbitmqstream sink pointing to the TLS port
Attempt to start Debezium Server
Connection fails because there is no way to configure TLS for the Stream client

**Feature request**
Could you please add TLS/mTLS support for the rabbitmqstream sink? The RabbitMQ Stream Java Client supports TLS via EnvironmentBuilder.tls().sslContext(...) with Netty's SslContextBuilder.
Ideally, the configuration would support PEM files directly, similar to the AMQP sink:
propertiesdebezium.sink.rabbitmqstream.tls.enabled=true
debezium.sink.rabbitmqstream.tls.trust.certificate.path=/path/to/ca.crt
debezium.sink.rabbitmqstream.tls.client.certificate.path=/path/to/client.crt
debezium.sink.rabbitmqstream.tls.client.key.path=/path/to/client.key
debezium.sink.rabbitmqstream.tls.hostname.verification=true
Thank you!

akshat08:
# Fix: TLS/mTLS support for the rabbitmqstream sink (debezium-server)

## Context

Fixes [debezium/dbz#1582](https://github.com/debezium/dbz/issues/1582): `RabbitMqStreamNativeChangeConsumer` (the `debezium.sink.type=rabbitmqstream` sink, in `debezium-server-rabbitmq`) only supports `tls.enable` — there's no way to trust a private/internal CA, and no way to present a client certificate for mTLS. The sibling AMQP sink (`rabbitmq`) gets this for free via `ConnectionFactoryConfigurator`, but that mechanism is specific to the AMQP client library and requires Java KeyStore files (PKCS12/JKS), not the reporter's PEM files.

## Config changes — `RabbitMqStreamNativeChangeConsumerConfig.java`

**Remove**: `TLS_SERVER_NAME` field (`tls.serverName`) and its getter — dead code, per above.

**Keep unchanged**: `TLS_ENABLE`.

**Add**, following the exact `Field.create(...)` pattern already used for `TLS_ENABLE`/`TLS_SERVER_NAME` (lines 74-87):

| Field | Property | Type | Notes |
|---|---|---|---|
| `KEY_STORE_TYPE` | `tls.keyStore.type` | STRING | Optional, no `Field`-level validation |
| `KEY_STORE_CERTIFICATE_FILE_PATH` | `tls.keyStore.certificateFilePath` | STRING | Optional. Used only when type=PEM |
| `KEY_STORE_KEY_FILE_PATH` | `tls.keyStore.keyFilePath` | STRING | Optional. Used only when type=PEM |
| `KEY_STORE_FILE_PATH` | `tls.keyStore.filePath` | STRING | Optional. Used only when type=PKCS12/JKS |
| `KEY_STORE_PASSWORD` | `tls.keyStore.password` | PASSWORD | Optional. Keystore password (PKCS12/JKS) or PEM key passphrase if encrypted |
| `TRUST_STORE_TYPE` | `tls.trustStore.type` | STRING | Optional, no `Field`-level validation |
| `TRUST_STORE_FILE_PATH` | `tls.trustStore.filePath` | STRING | Optional. Single file regardless of type |
| `TRUST_STORE_PASSWORD` | `tls.trustStore.password` | PASSWORD | Optional. PKCS12/JKS only |
| `TLS_VERIFY_HOSTNAME` | `tls.verify.hostname` | BOOLEAN | Default `false` |

Add corresponding getters, matching the existing `isTlsEnable()`/`getTlsServerName()` style.

**Add cross-field validation**, a new private `validateTlsOptions()` method, called at the end of `init(Configuration config)` (right after all the new fields above are assigned), checking:
- If `tls.keyStore.type` is set: value must be one of `PEM`/`PKCS12`/`JKS` (case-insensitive) — else throw `DebeziumException("Unsupported tls.keyStore.type '...'")`, matching Zerobus's exact message style for unsupported enum values.
- If type is `PEM`: `validatePair(certificateFilePath, "tls.keyStore.certificateFilePath", keyFilePath, "tls.keyStore.keyFilePath")`
- If type is `PKCS12`/`JKS`: `tls.keyStore.filePath` must be set — else throw clearly.
- Same three checks mirrored for `tls.trustStore.*` (type validity, then required file(s) for that type — trustStore only ever needs the single `filePath`, no pair check needed there).

## Consumer changes — `RabbitMqStreamNativeChangeConsumer.java`

**Remove** the dead SNI block and now-unused imports (`SNIHostName`, `SSLEngine`, `SSLParameters`, `Collections` if nothing else in the file needs it — verify at implementation time).

**Extract** the TLS block (current lines ~186-209) into a new package-private method, e.g.:
```java
SslContext buildSslContext(RabbitMqStreamNativeChangeConsumerConfig config) throws SSLException
```
This makes it independently unit-testable without a live broker connection. Because `RabbitMqStreamNativeChangeConsumerConfig`'s constructor already guarantees internal TLS consistency (per `validateTlsOptions()` above), this method can assume: if a type is set, its required file(s) are present. 

**In `connect()`**: call `buildSslContext(config)` inside the existing `if (config.isTlsEnable())` check; wrap any thrown exception in `DebeziumException` and let it propagate, failing startup rather than silently falling back to plaintext. This fixes the pre-existing silent-fallback risk for both old and new failure modes.

## Tests

- **`RabbitMqStreamNativeChangeConsumerConfigTest.java`** (new) 
- **`RabbitMqStreamNativeChangeConsumerTest.java`** (new) 
- Existing `RabbitMqStreamNativeSuperStreamIT.java` and the AMQP `RabbitMqStreamChangeConsumerTest.java` continue passing unmodified — confirms no regression to unrelated sink behavior.

## Verification

- `./mvnw test -pl debezium-server-rabbitmq -am` — 25/25 unit tests pass.
- Full reactor `./mvnw install` across `debezium-server` (all 25+ sink modules) — `debezium-server-rabbitmq` reports `SUCCESS`.

### Manual end-to-end TLS/mTLS verification against a real broker

Beyond the unit tests above (which verify `buildSslContext()` in isolation), I additionally verified the actual handshake against a real, TLS-configured RabbitMQ 4.2.0 broker — the same version reported in the original issue — using a genuine CA-signed certificate chain (CA → server cert, CA → client cert), not just standalone self-signed fixtures. Two scenarios, both a real connect + stream create/delete round-trip using the production `buildSslContext()` code path, not a mock:

- **Plain TLS**: broker configured with `ssl_options.verify = verify_peer` / `fail_if_no_peer_cert = false` (requests but does not require a client cert). Client configured with `tls.trustStore.type=PEM` only, no key store. Result: connects and completes a stream create/delete successfully, confirming the trust-only path works against a real server cert issued by a private CA.
- **Mutual TLS**: broker configured with `ssl_options.verify = verify_peer` / `fail_if_no_peer_cert = true` (client cert mandatory). Client configured with both `tls.trustStore.type=PEM` and `tls.keyStore.type=PEM` (client cert + key). Result: connects and completes a stream create/delete successfully, confirming the client correctly presents its identity and the broker accepts it.

Notable finding along the way, unrelated to this fix but worth documenting: RabbitMQ's Stream plugin TLS listener appears to default to `verify_peer` internally and will crash the plugin on startup (`{options, incompatible, [{verify, verify_peer}, {cacerts, undefined}]}`) if `ssl_options.cacertfile` isn't also set — even when the intent is a non-mutual, trust-only listener. Configuring `cacertfile` unconditionally alongside `fail_if_no_peer_cert=false` avoids this.

This verification setup (Testcontainers + a locally-generated CA hierarchy) was throwaway/local-only and is not part of this PR's diff.


debezium-builder (approved):
@akshat08 LGTM, thanks. Can be merged when CI passes.
Could you please also prepare a docs PR?

akshat08:
@debezium-builder Added a PR for documentation update as you asked: https://github.com/debezium/debezium/pull/7963

akshat08:
@Naros Let me know if there is something else required. I have added changes to the documentation and raised the PR too. I don't have permissions to merge the changes.

Naros:
Applied @akshat08, thanks! :rocket:

PR #314, opened 2026-09-04
debezium/dbz#2557 Reconcile the Redis memory estimate against the reported usage

### LINKED ISSUE #2557: Redis sink halts on a fabricated memory estimate: RedisMemoryThreshold accumulator is never reconciled
### What Debezium connector do you use and what version?

Debezium Server 3.0.0.Final, MySQL connector, Redis sink (Kafka-less).
Also verified against 3.6.1.Final, 3.6.2.Final, 3.7.0.Alpha1/Alpha2/Beta1.

### What is the connector configuration?

debezium.sink.type=redis
debezium.sink.redis.address=redis:6379
debezium.sink.redis.batch.size=1000
debezium.sink.redis.memory.limit.mb=0

38 captured tables across 4 MySQL databases -> 38 Redis streams.

### What is the captured database version and mode of deployment?

MySQL 8.0 replica, on-premises, Docker Compose.

### What behaviour do you expect?

The sink should back-pressure based on Redis's actual memory usage.

### What behaviour do you see?

The sink stops permanently, logging once a second:

    Sink memory threshold percentage was reached. Will retry;
    (estimated used memory size: 773.42 MB, maxmemory: 768.00 MB)
    Stopped consuming records!

At that moment Redis `INFO memory` reported **used_memory 1.5 MB**, rss 9.4 MB,
peak 1.5 MB. No Redis figure is anywhere near 773 MB. The snapshot froze at
table 1 of 36 and never resumed.

### Do you see the same behaviour using the latest released Debezium version?

Yes. `io/debezium/server/redis/RedisMemoryThreshold.class` is byte-for-byte
identical (`javap -p -c`, diffed) in 3.0.0.Final, 3.6.1.Final, 3.6.2.Final,
3.7.0.Alpha1, 3.7.0.Alpha2 and 3.7.0.Beta1.

### Implementation detail

`RedisMemoryThreshold` keeps a **static** accumulator that grows whenever Redis
reports an unchanged `used_memory`, and is only ever reset on the else branch:

```java
private static long accumulatedMemory;
private static long previouslyUsedMemory;

long delta = usedMemory - previouslyUsedMemory;
if (delta == 0) {
    accumulatedMemory += recordSize * numRecords;   // grows unbounded
} else {
    previouslyUsedMemory = usedMemory;
    accumulatedMemory    = recordSize * numRecords; // only reset here
}
long estimated = usedMemory + accumulatedMemory + bufferBytes;
if (estimated >= maximumMemory) { /* stop consuming */ }

```

`memoryTuple()` reads only `used_memory` and `maxmemory` from INFO.

Controlled test — change `maxmemory` alone, touch no data, restart nothing:

| maxmemory | result |
|---|---|
| 768 MB | estimate 773.42 MB, stopped |
| 2 GB   | snapshot resumed immediately |

The estimate did not change; only the threshold it was compared against did.
In our failure: 773.42 MB estimate − 1.43 MB actually used = **771.99 MB of
pure accumulator**.

Redis reports `used_memory` in coarse allocator jumps, so `delta == 0` is common,
especially with many small streams. The accumulator then climbs without ever
being reconciled against reality.

### Two documented properties that do not exist in the code

A binary scan of every shipped class finds zero matches for either:

- `debezium.sink.redis.maxlen` — never trims. Measured `XLEN` 115,409 with
  `maxlen=1000` set. `RedisClient.xadd()` takes no MAXLEN argument in any version.
- `debezium.sink.redis.memory.threshold.percentage` — setting it to `0` does
  **not** disable the check as documented; the check still fired 211 times.

The properties actually parsed are: `redis.batch.size`, `redis.buffer.fill.rate`,
`redis.memory.limit.mb`, `redis.message.format`, `redis.null.key`,
`redis.null.value` (plus `redis.skip.heartbeat.messages` from 3.6.1).

These look accepted because Debezium echoes the whole `debezium.sink.*` prefix
back in its startup log, known and unknown keys alike.

### Suggested fix

Reconcile the accumulator against observed usage rather than letting it grow on
`delta == 0` — e.g. clamp `estimated` to `usedMemory + bufferBytes` whenever a
real `used_memory` reading arrives, or drop the accumulator and use RSS.
Happy to open a PR.

Ansari4003:
Fixes https://github.com/debezium/dbz/issues/2557

### The problem

`RedisMemoryThreshold` adds every batch it sends to an accumulator while `used_memory` stays unchanged. Redis reports that figure in coarse allocator steps, so the same reading comes back across many batches, and the accumulator is never checked against what Redis actually reports. It grows until the estimate crosses `maxmemory` and the sink stops consuming for good while the database is nearly empty.

On the pipeline where this was found, the sink halted mid-snapshot with:

```
Sink memory threshold percentage was reached. Will retry;
(estimated used memory size: 773.42 MB, maxmemory: 768.00 MB)
```

logged once a second, forever, while Redis held **1.43 MB**. Changing only `maxmemory` from 768 MB to 2 GB — touching no data and restarting nothing — resumed the snapshot immediately: the estimate never changed, only the threshold it was compared against.

### The fix

Discard the unconfirmed part of the estimate once it alone would stop the sink while the reported usage still leaves room for another batch. The writes the accumulator claims are in flight did land in Redis, so a database really holding that much would report a different figure — an unchanged reading disproves the estimate.

Back-pressure still applies as soon as the reported usage itself leaves no headroom, which is what `RedisStreamMemoryThresholdIT` covers (it sets `maxmemory` to 1M and asserts the sink stops); that test passes.

The same commit also rolls back the last observed reading together with the accumulator when a batch is held back. Rolling back only the accumulator left the reading advanced, so the next call saw an unchanged reading and resumed accumulating from there.

### A note on the existing test

`testMemoryLimits` asserted that the sink stops after eight batches even though the fake client reports a constant 10 MB of 20 MB used — that is the behaviour described above, so the test had to be rewritten. It now asserts that consumption continues, and two new tests cover both sides: consumption continues while the reported usage stays constant, and still stops once the reported usage itself leaves no headroom. The first of those fails on the unpatched code; the second guards against over-correcting and passes either way.

### The second commit

The three counters were `static` while every other field in the class is not, so one `RedisMemoryThreshold` saw the estimate built up by another and the state outlived each test, which made the accumulator behaviour impossible to exercise in isolation. It is a separate commit — happy to drop it and send it separately if you would rather keep this PR to the fix alone.

### Verification

- `RedisMemoryThresholdTest` — 5 tests pass; `testConsumptionContinuesWhileReportedMemoryStaysConstant` fails on the unpatched code with "Sink stopped consuming at batch 8 while Redis still reported 10MB of 20MB used"
- `RedisStreamMemoryThresholdIT` — passes against a real Redis and Postgres
- Checkstyle, formatter and import-order checks clean
- Each commit builds and passes on its own

The class is byte-for-byte identical in 3.0.0.Final, 3.6.1.Final, 3.6.2.Final, 3.7.0.Alpha1, 3.7.0.Alpha2 and 3.7.0.Beta1, so this is not a regression from a recent release.


debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
```diff
@@ -106,18 +107,60 @@ private static void setField(Object target, Class<?> declaringClass, String fiel
...
         redisMemoryThreshold.setRedisClient(new RedisClientImpl(_5MB, _20MB));
         Assertions.assertTrue(redisMemoryThreshold.checkMemory(RECORD_SIZE, BUFFER_SIZE, RATE_PER_SECOND));
     }
 
+    /**
+     * Redis reports 'used_memory' in coarse allocator steps, so the same reading is returned across many
+     * batches. The accumulated estimate must not grow past what Redis reports for ever: half the database
+     * is free the whole time here, so the sink has to keep consuming.
+     */
+    @Test
```
Naros:
```suggestion
    @Test
    @FixFor("debezium/dbz#2557")
```
We should add these for all the new test methods.

debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
```diff
@@ -106,18 +107,60 @@ private static void setField(Object target, Class<?> declaringClass, String fiel
...
-        Configuration config = Configuration.from(Collect.hashMapOf("debezium.sink.redis.address", "localhost",
-                "debezium.sink.redis.rate.per.second", RATE_PER_SECOND));
-        RedisMemoryThreshold redisMemoryThreshold = new RedisMemoryThreshold(new RedisClientImpl(_10MB, _20MB),
-                new RedisStreamChangeConsumerConfig(config));
+        RedisMemoryThreshold redisMemoryThreshold = memoryThreshold(new RedisClientImpl(_10MB, _20MB));
         for (int i = 0; i < 8; i++) {
             Assertions.assertTrue(redisMemoryThreshold.checkMemory(RECORD_SIZE, BUFFER_SIZE, RATE_PER_SECOND));
         }
-        Assertions.assertFalse(redisMemoryThreshold.checkMemory(RECORD_SIZE, BUFFER_SIZE, RATE_PER_SECOND));
+        Assertions.assertTrue(redisMemoryThreshold.checkMemory(RECORD_SIZE, BUFFER_SIZE, RATE_PER_SECOND));
```
Naros, line 114: Doesn't this change now only assert the usage guard and not the accumulator?

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -72,22 +72,49 @@ public boolean checkMemory(long extraMemory, int bufferSize, int bufferFillRate)
...
         if (estimatedUsedMemory >= maximumMemory) {
             LOGGER.info(
                     "Sink memory threshold percentage was reached. Will retry; "
                             + "(estimated used memory size: {}, maxmemory: {}). Total Processed Records: {}",
                     getSizeInHumanReadableFormat(estimatedUsedMemory), getSizeInHumanReadableFormat(maximumMemory), totalProcessed);
+            // The batch was not sent, so neither the accumulator nor the last observed reading may
+            // advance. Rolling back only the accumulator would leave 'previouslyUsedMemory' updated,
+            // making the next call see an unchanged reading and resume accumulating from there.
             accumulatedMemory = prevAccumulatedMemory;
+            previouslyUsedMemory = prevPreviouslyUsedMemory;
```
Naros: I don't believe there is any test that asserts this previouslyUsedMemory rollback.

debezium-server-redis/src/test/java/io/debezium/server/redis/RedisMemoryThresholdTest.java
Ansari4003: You're right. testMemoryLimits now covers only the usage guard.

The accumulator behaviour moved into two dedicated tests: testConsumptionContinuesWhileReportedMemoryStaysConstant (the estimate must not grow past what Redis reports) and testConsumptionStopsWhenReportedMemoryLeavesNoHeadroom (back-pressure must still apply when the reported usage genuinely leaves no room).

The original assertion asserted the bug itself - it required the sink to stop after 8 batches while the fake client reported a constant 10MB of 20MB used. I changed it rather than deleting it so the change stays visible in the diff, but I'm happy to restructure if you'd rather keep the accumulator assertions in this test.

debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
Ansari4003: You're right, and it goes further than missing coverage - I went looking for a test and couldn't write one.

I simulated checkMemory's arithmetic over three million randomised sequences and could not find a case where rolling back previouslyUsedMemory changes the return value. At the rejection point previouslyUsedMemory always already equals usedMemory - the diff==0 branch leaves it equal, and both the else branch and the reconciliation assign it - so the rollback only restores a value that is then either equal anyway or reconciled on the next call. The internal state can differ for one call when Redis repeats the same reading, but the reconciliation guard collapses that path back to the same accumulator, so the decision never differs.

So the line was unobservable and the comment above it described a failure mode that doesn't occur. I've dropped both - the accumulator rollback is the one that matters, and it stays. Force-pushed.

PR #318, opened 2026-09-22
feat(eventhubs): add Azure managed identity authentication for Event Hubs sink

abhi-roop:
**## Summary**

Adds support for Azure Entra ID authentication using `DefaultAzureCredential`, enabling secretless deployments of the Event Hubs sink on Azure Container Apps, AKS, VMs, and other Azure-hosted environments.

Fixes #2647.

## New configuration

| Property | Default | Description |
|----------|---------|-------------|
| `debezium.sink.eventhubs.authmode` | `connection-string` | Set to `default-azure-credential` to use Entra ID |
| `debezium.sink.eventhubs.fullyqualifiednamespace` | — | Required when `authmode=default-azure-credential` (e.g. `myns.servicebus.windows.net`) |

The existing `connectionstring` configuration remains supported and is still the default, so there is no breaking change.

## Notes

- I used `authmode` and `fullyqualifiednamespace` to follow the existing flat naming convention in `EventHubsChangeConsumerConfig` (e.g. `connectionstring`, `hubname`, `partitionid`). Happy to switch to `auth.mode` / `fully-qualified-namespace` if preferred.
- A user-assigned managed identity can be selected via the standard `AZURE_CLIENT_ID` environment variable consumed by Azure Identity.
- The workload identity needs the **Azure Event Hubs Data Sender** role at namespace or hub scope.

## Testing

- `mvn -pl debezium-server-eventhubs -am clean install` passes locally, including integration tests via Testcontainers.
- Added unit tests covering config parsing and validation of the new auth mode.


debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
```diff
@@ -83,21 +85,41 @@ void connect() {
         final Config mpConfig = ConfigProvider.getConfig();
 
         // Load configuration
-        io.debezium.config.Configuration configuration = io.debezium.config.Configuration.from(getConfigSubset(mpConfig, PROP_PREFIX));
+        io.debezium.config.Configuration configuration = io.debezium.config.Configuration
+                .from(getConfigSubset(mpConfig, PROP_PREFIX));
```
Naros, line 89: The PR includes about 6 spurious reformatting changes that are unnecessary and add noise. Could we reduce the changes here to only what's applicable to the feature?

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumerConfig.java
```diff
@@ -29,6 +29,21 @@ public class EventHubsChangeConsumerConfig {
...
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("Authentication mode for Event Hubs. Use 'connection-string' (default) or 'default-azure-credential'.");
+
+    public static final Field FULLY_QUALIFIED_NAMESPACE = Field.create("fullyqualifiednamespace")
+            .withDisplayName("Fully Qualified Namespace")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("Fully qualified Event Hubs namespace (e.g. <namespace>.servicebus.windows.net). Required when authmode is 'default-azure-credential'.");
```
Naros, line 45: If you look at the Fluss and ZeroBus modules, you'll notice that for fields that have a finite set of values, such as what you've outlined here, we use an `enum` that implements `EnumeratedValue` and then, inside the init, we have the `enum` use its parse method. This has two benefits:

1. Can throw an exception if the field value parses to `null` or an invalid value.
2. Can use enum value sets throughout rather than string comparisons.

Let's adopt that precedent here for consistency.

debezium-server-eventhubs/src/main/java/io/debezium/server/eventhubs/EventHubsChangeConsumer.java
```diff
@@ -83,21 +85,41 @@ void connect() {
...
-            dynamicPartitionRoutingStrategy = DynamicPartitionRoutingStrategy.fromString(config.getDynamicPartitionRouting());
+            dynamicPartitionRoutingStrategy = DynamicPartitionRoutingStrategy
+                    .fromString(config.getDynamicPartitionRouting());
         }
         hashMessageFunction = Optional.ofNullable(config.getHashMessageKeyFunction()).map(HashFunction::fromString);
 
-        String finalConnectionString = String.format(CONNECTION_STRING_FORMAT, config.getConnectionString(), config.getEventHubName());
+        if ("default-azure-credential".equals(config.getAuthMode())) {
+            String namespace = config.getFullyQualifiedNamespace();
+            if (namespace == null || namespace.isEmpty()) {
```
Naros, line 102:
```suggestion
            if (Strings.isNullOrEmpty(namespace)) {
```
`debezium-util` includes a full suite of utility classes, like `Strings`. We prefer to use those for simplicity.

Naros (changes requested):
Hi @abhi-roop I've left a few inline comments.

One point about the PR description, you mention the use of `AZURE_CLIENT_ID`, but I don't recall that being documented in the docs PR that accompanies this change. We should probably be sure to include that somehow for users.

abhi-roop:
Hi @Naros, thanks for the review. I've addressed the feedback:

- Removed the spurious reformatting — restored the file from upstream and re-applied only the feature changes.
- Switched `authmode` to an `AuthMode` enum implementing `EnumeratedValue`, following the Fluss `PrimaryKeyMode` precedent.
- Used `Strings.isNullOrEmpty` for the namespace check.

Also updated the docs PR #8082 to mention `AZURE_CLIENT_ID` for user-assigned managed identities, as you noted. Ready for another look when you have time.

